PR #24, opened 2020-07-20
Create `category-scholar-cn`

IceCodeNew:
0. Add apabi.com
1. Add cnthinkers.com
2. Move cqvip.com

### DIFF
--- a/data/category-scholar-cn
+++ b/data/category-scholar-cn
@@ -1,0 +1,7 @@
+# This list contains scholarly domains that have access point in China Mainland.
+
+include:cnki
+
+apabi.com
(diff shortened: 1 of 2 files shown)


EpLiar:
If someone's config file use `geosite:cnki`, Core will crash when they update dlc file.

So include `cnki` into `category-scholar-cn` is better than remove cnki.

IceCodeNew:
> If someone's config file use `geosite:cnki`, Core will crash when they update dlc file.
> 
> So include `cnki` into `category-scholar-cn` is better than remove cnki.

Fixed.

PR #84, opened 2020-07-27
Feat: export lists in plaintext format

Loyalsoldier:


Loyalsoldier:
Flatten, generate and export lists that you care in plaintext format. And push text files to `release` branch.

Already tested in my repo: https://github.com/Loyalsoldier/domain-list-community/runs/915597417
Generated lists can be found here: https://github.com/Loyalsoldier/domain-list-community/tree/release

main.go
```diff
@@ -319,6 +344,24 @@ func main() {
...
+					fileName := filepath.Join(dir, exportedListName)
+					_, err := os.Stat(fileName)
+					if err == nil || os.IsExist(err) {
+						existList = append(existList, exportedListName)
+					} else {
+						fmt.Printf("'%s' list does not exist in '%s' directory.\n", exportedListName, dir)
+					}
+				}
+			}
+		}
```
iochen, line 364: If happens that when ranging, the first `refName` is the one you want to export, then it cannot be exported.
It should be finished before ranging.

PR #85, opened 2020-07-27
geolocation-cn: remove Google domain

akiirui:
google: Add attributes @cn to certain domain that available in China mainland

Ref: https://github.com/v2ray/domain-list-community/issues/487#issuecomment-664522925

### DIFF
--- a/data/geolocation-cn
+++ b/data/geolocation-cn
@@ -17,12 +17,6 @@ full:cdn.ampproject.org
 ## jsDelivr
 full:cdn.jsdelivr.net
 
-## Google
-recaptcha.net
-full:fonts.googleapis.com
-full:fonts.gstatic.com
-full:safebrowsing.googleapis.com
-
 ## 网宿科技
 ourdvsss.com
 
--- a/data/google
+++ b/data/google
@@ -11,6 +11,12 @@ include:polymer
 include:v8
 include:youtube
 
+# Available on China mainland
+recaptcha.net @cn
+full:fonts.googleapis.com @cn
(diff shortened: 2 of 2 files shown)


Robot-DaneelOlivaw:
You speak out my mind. It's more elegant to use `@cn` than listing all those domains in `geolocation-cn`. We should treat other similar domains in the same way.

By the way, it would be great if we could use attributes directly in the list. Then `include:google@cn` can be added back to `geolocation-cn`.

akiirui:
> You speak out my mind. It's more elegant to use `@cn` than listing all those domains in `geolocation-cn`. We should treat other similar domains in the same way.
> 
> By the way, it would be great if we could use attributes directly in the list. Then `include:google@cn` can be added back to `geolocation-cn` .

Oh please don't add back to `geolocation-cn`. Because if user using DNS that outside China, received IP cannot directly visit.

Robot-DaneelOlivaw:
> Oh please don't add back to `geolocation-cn`. Because if user using DNS that outside China, received IP cannot directly access.

So people don't use Chinese DNS for `geolocation-cn`? In this case, putting `geolocation-!cn` in front can override these domains.

With that said, users have to be familiar with domain lists to correctly configure. Leaving it out might be safer.

Anyway, it doesn't affect this PR.

> Maybe we can create a new category `category-available-cn` similar to `category-ads`. Integrate all domain from GFW sites but available in China.

It's a great idea to manage these domains altogether. We'll have to see if this function will be deployed though. (And I might choose a different name for it.)

PR #89, opened 2020-07-28
Add apple & google direct list

Loyalsoldier:


IceCodeNew:
> I don't quite like the idea of `@direct`, mostly because of the name. It's implying that these domains shouldn't be proxied.

附议。

Loyalsoldier:
I would like to implement this first: https://github.com/v2fly/domain-list-community/pull/88#issuecomment-664765447

Mark this PR as draft for now.

Robot-DaneelOlivaw:
The lists are worth adding in my opinion. We just need to find a proper name for them. I reckon `@cn` would be fine. Personally, I prefer adding them directly in `data/apple` and `data/google` instead of creating new lists.

Also, are we going to maintain this kind of lists for every company? Or just Apple and Google?

PR #115, opened 2020-08-06
Update homebrew

EpLiar:
image

I noticed this domain in my daily use of Homebrew and made sure to commit after I encountered it multiple times.

### DIFF
--- a/data/homebrew
+++ b/data/homebrew
@@ -1,1 +1,3 @@
 brew.sh
+
+full:homebrew.bintray.com
--- a/data/jfrog
+++ b/data/jfrog
@@ -1,3 +1,5 @@
 jfrog.com
 jfrog.org
 bintray.com
+
+full:d29vzk4ow07wi7.cloudfront.net


Robot-DaneelOlivaw:
I've dug in a bit and this CDN seems to be related to [JFrog Bintray](https://bintray.com/), as it's frequently used by Maven, Gradle, Homebrew and more, with the domains `jcenter.bintray.com` or `homebrew.bintray.com`. Bintray probably provides software distribution service for them.

So you might need to add it into `jfrog`. Anyway, nice catch.

IceCodeNew:
> Bintray probably provides software distribution service for them.`jcenter.bintray.com``homebrew.bintray.com`
> 
> So you might need to add it into

Agreed,

EpLiar:
Oh, I see.

But if I add `full:homebrew.bintray.com` into `homebrew`, it will redirect to CloudFront when visit.

So it looks like this PR intends to do something that needs to be changed?

Robot-DaneelOlivaw:
`full:homebrew.bintray.com` in `homebrew` and `full:d29vzk4ow07wi7.cloudfront.net` in `jfrog` will do.

PR #157, opened 2020-08-17
Add speedtest keyword to ookla-speedtest.

RevNelson:
The number of possible server matches is unrealistic to try to match with, but it seems that during the test, partial files are downloaded from something like speedtest.closest-server.com.

From the README, it seems the line should read `keyword:speedtest` but I noticed no other keywords were prefaced with that, so I omitted it. Let me know if you want me to change it.

### DIFF
--- a/data/ookla-speedtest
+++ b/data/ookla-speedtest
@@ -5,6 +5,7 @@ ekahau.com
 ookla.com
 ooklaserver.net
 pingtest.net
+speedtest
 speedtest.co
 speedtest.net
 speedtestcustom.com


Loyalsoldier:
It's pointless and will match irrelevant domains. There is no effective way to match Speedtest's domains since no patterns in it.

RevNelson:
That's unfortunate. Without some solution, there's no real way to set `speedtest` rules and quickly switch between servers. Do you know of a better site/method?

RevNelson:
> First, you will need `keyword:` prefix, or you'll got a domain rule.
> 
> And I don't think keyword `speedtest` is only used by Ookla speedtest.

You're right that it doesn't seem specific enough. I have manually entered it in my rules and haven't encountered any ill-effects, but I concede it could eventually snag something unwanted.

As for the `keyword:` prefix, searching the entire directory shows only `main.go` and `README.md` as containing that string. I also agree that, based on the README, it should have it. I only left it out because I don't see it anywhere else. For example in the `google` file, there seem to be many keywords without the prefix.

Robot-DaneelOlivaw:
> For example in the `google` file, there seem to be many keywords without the prefix.

Those without prefix are actually `domain:`, which are TLDs of Google. IIRC, `domain:goog` matches `pki.goog`, `ocsp.pki.goog` but not `goog.com` or `google.com`, while `keyword:goog` does.

So `keyword:speedtest` is fairly radical. It probably matches not only domains of Ookla Speedtest but any other speedtest services as well. In this case, I would suggest users adding their own rules instead of modifying the whole list.

PR #189, opened 2020-09-02
Update bloomberg

EpLiar:


data/bloomberg
```diff
@@ -1,26 +1,48 @@
 # All .bloomberg domains
 bloomberg
 
+bbgevent.com
 bbhub.io
 bbthat.com
 bgov.com
-bloomberg.cn
+bloomberg.cn @cn
```
Robot-DaneelOlivaw, line 8: `bloomberg.cn` doesn't have access point in mainland China.

blessingCN: if I ping `bloomberg.cn` , I will get a US IP and it was blocked in mainland. (dns problem perhaps)

EpLiar: > `bloomberg.cn` doesn't have access point in mainland China.

Done.

Robot-DaneelOlivaw (approved):

PR #201, opened 2020-09-08
Update cloudflare

Loyalsoldier:


Loyalsoldier:
It seems it's not a Cloudflare domain.

PR #215, opened 2020-09-19
generate gfwlist.txt

kslr:
domain 转换成 ||
full 转换成 |http:// 和 |https://
keyword 直接输出
regex 转换成 /....../

TIP git squash

### DIFF
--- a/main.go
+++ b/main.go
@@ -59,39 +59,6 @@ func (l *ParsedList) toPlainText(listName string) error {
 	return nil
 }
 
-func (l *ParsedList) toGfwList() error {
-	var entryBytes []byte
-	entryBytes = append(entryBytes, []byte("[AutoProxy 0.2.9]\n")...)
-	for _, entry := range l.Entry {
-		switch entry.Type {
-		case "domain":
-			entryBytes = append(entryBytes, []byte("||"+entry.Value+"\n")...)
-		case "full":
-			entryBytes = append(entryBytes, []byte("|http://"+entry.Value+"\n")...)
-			entryBytes = append(entryBytes, []byte("|https://"+entry.Value+"\n")...)
-		case "keyword":
-			entryBytes = append(entryBytes, []byte(entry.Type+entry.Value+"\n")...)
-		case "regexp":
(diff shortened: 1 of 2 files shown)


Loyalsoldier:
I think it's better to move this code block outside the loop range:

```go
if refName == "GEOLOCATION-!CN" {
  if err := pl.toGfwList(); err != nil {
    fmt.Println("Failed: ", err)
  }
}
```

And then rename the func `exportPlainTextList` to `exportList`

main.go
```diff
@@ -58,6 +59,39 @@ func (l *ParsedList) toPlainText(listName string) error {
...
+	entryBytes = append(entryBytes, []byte("[AutoProxy 0.2.9]\n")...)
+	for _, entry := range l.Entry {
+		switch entry.Type {
+		case "domain":
+			entryBytes = append(entryBytes, []byte("||"+entry.Value+"\n")...)
+		case "full":
+			entryBytes = append(entryBytes, []byte("|http://"+entry.Value+"\n")...)
+			entryBytes = append(entryBytes, []byte("|https://"+entry.Value+"\n")...)
+		case "keyword":
+			entryBytes = append(entryBytes, []byte(entry.Type+entry.Value+"\n")...)
```
studentmain, line 73: []byte(entry.Value+"\n")

xiaokangwang, line 73: Sorry, I am not an expert at ABP rules, could you please explain why you need a `entry.Type`(which should always === "keyword") at the beginning of rule?

kslr:
> I think it's better to move this code block outside the for loop:
> 
> ```go
> if refName == "GEOLOCATION-!CN" {
>   if err := pl.toGfwList(); err != nil {
>     fmt.Println("Failed: ", err)
>   }
> }
> ```
> 
> And then rename the func `exportPlainTextList` to `exportList`.
> 
> The following code
> 
> ```go
> case "keyword":
>     entryBytes = append(entryBytes, []byte(entry.Type+entry.Value+"\n")...)
> ```
> 
> is missing a string separator between type and value. I wonder if GFWList supports keyword?

Fix

studentmain (approved):

Loyalsoldier (approved):

PR #229, opened 2020-09-26
Update private

Loyalsoldier:
As per https://github.com/v2fly/v2ray-core/pull/228

### DIFF
--- a/data/private
+++ b/data/private
@@ -5,9 +5,19 @@ local
 localhost
 
 # Also defined in rfc2606
-# test
-# example
-# invalid
+test
+example
+invalid
+
+# References:
+# https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml
+# https://unix.stackexchange.com/questions/92441/whats-the-difference-between-local-home-and-lan
+localdomain
+lan
+home.arpa
+
+# Dotless domains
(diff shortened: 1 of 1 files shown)


EpLiar (approved):

data/private
```diff
@@ -5,9 +5,19 @@ local
...
-# example
-# invalid
+test
+example
+invalid
+
+# References:
+# https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml
+# https://unix.stackexchange.com/questions/92441/whats-the-difference-between-local-home-and-lan
+localdomain
```
Vigilans, line 15: Isn't '.local' domain included?

Loyalsoldier: It is at the second line of the file.

data/private
```diff
@@ -5,9 +5,19 @@ local
...
+lan
+home.arpa
+
+# Dotless domains
+regexp:^[^.]+$
 
 # OpenWrt LuCI Web Interface
 openwrt.lan
```
Robot-DaneelOlivaw, line 23: Since we have `lan` now, `openwrt.lan` can be removed.

Loyalsoldier: https://github.com/v2fly/domain-list-community/pull/230

PR #236, opened 2020-09-28
Refine & exclude rules with @cn attribute from gfwlist.txt

Loyalsoldier:


Loyalsoldier:
The generated file header now will be like this:

```
[AutoProxy 0.2.9]
! Last Modified: 2020-09-29 00:00:00.955596 +0000 UTC m=+0.119235800
! Expires: 24h
! HomePage: https://github.com/v2fly/domain-list-community
! GitHub URL: https://raw.githubusercontent.com/v2fly/domain-list-community/release/gfwlist.txt
! jsdelivr URL: https://cdn.jsdelivr.net/gh/v2fly/domain-list-community@release/gfwlist.txt
```

studentmain (approved):
ss-windows now use geosite database directly and has no gfwlist support, so it won't solve the problem.

KevinZonda (approved):

Loyalsoldier:
> ss-windows now use geosite database directly and has no gfwlist support, so it won't solve the problem.

emm……

PR #241, opened 2020-10-01
update porn

ajiou:


data/category-porn
```diff
@@ -112,6 +123,15 @@ tssp.best
 tubetubetube.com
 upjav.cc
 uue63.com
+uug22.com
```
studentmain, line 126: will regex works here? maybe extract to another file?

Loyalsoldier:
If several domains are belong to one brand, I think it's better to organize them as a separate file named after the brand, then include it into the `category-porn` list like `include:pornhub`.

Loyalsoldier:
`t66y.com` has already been in `caoliu` list. Before adding a domain, first search if it's in the project.

IceCodeNew (approved):

IceCodeNew (approved):

studentmain (approved):

PR #247, opened 2020-10-11
Update private

database64128:


EpLiar (approved):

Loyalsoldier:
`include:filename@attr` doesn't work.

Loyalsoldier:
I think we should rewrite the build process and find a better way in several days. So draft it for now.

kslr:
What would you do

Loyalsoldier:
> What would you do

I'll rewrite the build process and add some features.

database64128:
I think we should merge the PR first. We can make changes after the new build process is deployed.

Loyalsoldier:
I think we should move all `@private` domains to the `private` file.

database64128:
> I think we should move all `@private` domains to the `private` file.

Agreed. But you forgot to add `openwrt.lan` to `private` in 16abfd4. 😅

Loyalsoldier:
> Agreed. But you forgot to add `openwrt.lan` to `private` in [16abfd4](https://github.com/v2fly/domain-list-community/commit/16abfd4399b25f665f051ac337dfe9edec8e5fe4). 😅

`lan` TLD is already in `private` file.

PR #258, opened 2020-10-26
Update category-ads

cky1201:
Rules updated from current chan.sankakucomplex.com ad providers.

### DIFF
--- a/data/category-ads
+++ b/data/category-ads
@@ -64,8 +64,16 @@ ad.wang502.com
 adbutter.net
 ads.trafficjunky.net
 adtechus.com
+adxprtz.com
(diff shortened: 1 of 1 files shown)


Loyalsoldier:
I think `cretgate.com` is not an ad domain. It's unnecessary to add `domain:` before a domain.

KevinZonda:
> I think `cretgate.com` is not an ad domain. It's unnecessary to add `domain:` before a domain.

I guess it is. If visit <cretgate.com>, website will redirect to an R-18 like ads. And it is so similar to <certgate.com>

Loyalsoldier (approved):

PR #262, opened 2020-11-02
🎵 Add `spotify-ads`

database64128:


data/spotify-ads
```diff
@@ -0,0 +1,3 @@
+full:adeventtracker.spotify.com @ads
+full:adstudio-assets.scdn.co @ads
+full:bloodhound.spotify.com @ads
```
IceCodeNew, line 3: https://github.com/x0uid/SpotifyAdBlock/issues/15#issuecomment-647056082

你自己验证过这个域名的用途吗？

IceCodeNew (changes requested):
我以前在为 https://github.com/DNSCrypt/dnscrypt-proxy 做贡献的时候接触过 DNS 层面的 spotify 广告规则，当时的规则更新方最后做了结论是这件事在 DNS 层面上是解决不了的
我一时找不到当时的材料，不过确实我认为这个 PR 可以再完善一点再合并。现在这三条规则应该是不足以包括 spotify 使用的广告服务的。

data/spotify-ads
database64128: You only see requests to `bloodhound.spotify.com` when you block access to ad sites.

database64128:
I have checked all domain names my Spotify desktop client connect to during daily usage. The `spotify-ads` was never meant to be used on its own, but as a part of `category-ads-all`. That’s why I didn’t include many other ad service domain names that are already in the list.

database64128:
My understanding is, a `*-ads` list should only contain in-house ad service domain names by the service provider. At least that’s what I observed from other `*-ads` lists.

IceCodeNew:
Then, we are good to go.

PR #276, opened 2020-11-22
Create infinitynewtab

Vixb1122:


Loyalsoldier:
I don't think these two domains are belong to the same entity. You can add them into `geolocation-cn` list by alphabetical order.

Vixb1122:
over

Loyalsoldier (approved):
Thanks for your work 🎉

PR #277, opened 2020-11-22
Update geolocation-cn

Vixb1122:


Loyalsoldier:
There is No access point for `snipaste.com` in China mainland. It's inappropriate to add it to `geolocation-cn` list.

Vixb1122:
`zh.snipaste.com` instead

Loyalsoldier:
> `zh.snipaste.com` instead

No access point in China mainland either:

image

Vixb1122:
all right！

PR #278, opened 2020-11-22
Update microsoft

Vixb1122:


Vixb1122:
Image from Microsoft Store App use 'store-images.s-microsoft.com'

PR #294, opened 2020-12-06
Move videojj.com back to geolocation-cn

Loyalsoldier:


Loyalsoldier:
AFAIK, videojj.com is not belong to zhanqi.tv

parasomn1a (approved):

IceCodeNew:
Good catch.

PR #347, opened 2021-01-06
Refactor category-media

Loyalsoldier:
Remove lists that only have one domain and move those domains to `category-media`.


IceCodeNew (approved):

jslll135:
Thanks! This change breaks my config. Why make such a breaking change without any warning?

PR #381, opened 2021-01-28
Update apple

Loyalsoldier:


Loyalsoldier:
ref: https://raw.githubusercontent.com/felixonmars/dnsmasq-china-list/master/apple.china.conf

PR #401, opened 2021-02-10
Add hkbn and smartone.

LiJingqing:


Loyalsoldier:
You should also add the domains that are in their TLS certification.

PR #418, opened 2021-02-18
few Google Earth used domain is no longer available in China

chenshaoju:
Tested domain:
kh.google.com
khm0.google.com

Test result:

image

ref: https://github.com/shadowsocks/shadowsocks-windows/issues/3103


kslr:
> regexp:^khm([0-3]|db)?\.google(apis)?\.com$ @cn

This won't work either.

kslr (approved):

PR #466, opened 2021-03-14
Fix: reorganize tver

Loyalsoldier:


Loyalsoldier:
`tver.co.jp` is the website of the company, and you can find other domains in the page `https://tver.co.jp/works/`

IceCodeNew (approved):

PR #467, opened 2021-03-14
Update category-ads

cky1201:
Added some domains found for ads on chan.sankakucomplex.com and rule34.xxx.
Note that I haven't been able to verify if those domains are mere image hosting platform. But due to its nature as providing ads, I suggest to add them here.

### DIFF
--- a/data/category-ads
+++ b/data/category-ads
@@ -59,7 +59,6 @@ include:zynga-ads
 a.adtng.com
 aaxads.com
 addthisedge.com
-addthis.com
 adtrue.com
 ad-srv.net
 ad.api.moji.com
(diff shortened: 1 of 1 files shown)


data/category-ads
```diff
@@ -58,6 +58,9 @@ include:zynga-ads
 51y5.net
 a.adtng.com
 aaxads.com
+addthisedge.com
+addthis.com
```
Loyalsoldier, line 62: This is not ad domain:

```
addthis.com
```

I can't confirm the following:

```
otaserve.net
trwl1.com
```

Loyalsoldier:
I remove the domains that I can't confirm.

PR #503, opened 2021-04-14
🍏 Remove `@cn` from `apps.apple.com`

database64128:
The Chinese CDN of `apps.apple.com` returns empty response on many apps. For example, you can try `curl https://apps.apple.com/us/app/leaf-lightweight-proxy/id1534109007` and compare the results.

### DIFF
--- a/data/apple
+++ b/data/apple
@@ -797,7 +797,6 @@ full:adcdownload.apple.com.akadns.net @cn
 full:adcdownload.apple.com @cn
 full:appldnld.apple.com @cn
 full:appldnld.g.aaplimg.com @cn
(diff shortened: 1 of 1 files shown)


Loyalsoldier:
I think that's a restriction on purpose. If you run `curl https://apps.apple.com/cn/app/qq/id444934666`, nothing gets wrong.

database64128:
> I think that's a restriction on purpose.

That’s why `@cn` must be removed from the sub domain.

IceCodeNew (approved):

IceCodeNew:
I do agree.

PR #528, opened 2021-05-19
Update tiktok

funray:
amemv.com is not blocked in Mainland China.

### DIFF
--- a/data/tiktok
+++ b/data/tiktok
@@ -1,6 +1,3 @@
-# It is not blocked in Mainland China
-#amemv.com
-
 muscdn.com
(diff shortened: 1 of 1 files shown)


data/tiktok
```diff
@@ -1,4 +1,6 @@
-amemv.com
+# It is not blocked in Mainland China
+#amemv.com
+
```
Loyalsoldier, line 3:
```suggestion
```

funray:
Enable the proxy of amemv.com will make DouYin display the location as the location of the VPS.

PR #530, opened 2021-05-20
Add more bytedance list

sayomelu:


data/lark
```diff
@@ -0,0 +1,7 @@
+# Lark
+larksuite.com
+larksuitecdn.com
+
+# Feishu
+feishu.cn
+feishucdn.com
```
kslr, line 7: "feishu" belong "lark" ?

sayomelu: https://en.wikipedia.org/wiki/Lark_(software) different name in different region

kslr (approved):
Ok, Thanks for your work!

PR #583, opened 2021-07-18
Add curseforge, planetminecraft

foreversam:
These sites do not belong to Mojang, but may be useful for Minecraft mods players.


Loyalsoldier:
I have made some modifications.

PR #585, opened 2021-07-22
added `linetv`; updated `category-media`

phlinhng:
1. Although `linetv` seems to use a single domain currently, creating a separated field for it is still necessary for detailed routing.
2. Included `linetv` in `line` for forward compatibility
3. Added `linetv` and `primevideo` to `category-media`

### DIFF
--- a/data/category-media
+++ b/data/category-media
@@ -27,6 +27,7 @@ include:huffpost
 include:infowars
 include:insider
 include:kanzhongguo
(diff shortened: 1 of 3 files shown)


Loyalsoldier:
`category-media` is for news and press.
`category-entertainment` is for games, music, podcasts, videos.
So I think this is unnecessary, thanks for contribution though.

PR #597, opened 2021-08-05
Update roku

tritu:


Loyalsoldier:
vudu is not owned by roku

tritu:
vudu is hardcoded in all Roku TV. maybe it's time to create another one. close this PR now.

PR #611, opened 2021-09-05
增加category-analytics

degfw:


IceCodeNew:
If these domains are sorted out by others or referring to internet sources, please label them out.

degfw:
@IceCodeNew I'm sorry for so long to reply to you.

For these domains, `in.appcenter.ms` and `app-measurement.com` are found from my device, and others were collecting from GitHub. 

However, I couldn't re-find them all at once, so I can't referring their sources all here.

IceCodeNew:
> @IceCodeNew I'm sorry for so long to reply to you.
> 
> For these domains, `in.appcenter.ms` and `app-measurement.com` are found from my device, and others were collecting from GitHub.
> 
> However, I couldn't re-find them all at once, so I can't referring their sources all here.

Then you should only submit domains that you are sure they are utilizing ONLY for tracking.

data/category-analytics
```diff
@@ -0,0 +1,30 @@
+# List of all domains being used in log collection or monitor.
+
+include:category-ads
```
IceCodeNew, line 3: should not include `ads` as they are completely different use cases.

data/category-analytics
```diff
@@ -0,0 +1,30 @@
...
+
+aax-us-east.amazon-adsystem.com
+alb.reddit.com
+analytics.brave.com
+analytics.map.qq.com
+analytics.plex.tv
+app-measurement.com
+be-nl.norton.com
+c.s-microsoft.com
```
IceCodeNew: If providing the referring sources turns to be difficult and painful, you could just concatenate all the domain lists that you have accessed during doing this PR, then give me the all-in-one list for me to do the review.

IceCodeNew (changes requested):
If providing the referring sources turns to be difficult and painful, you could just concatenate all the domain lists that you have accessed during doing this PR, then give me the all-in-one list for me to do the review.

degfw:
The only domains I can find from my device are the three left, and they are used within Wechat, Clash, and Google's app.

IceCodeNew:
> The only domains I can find from my device are the three left, and they are used within Wechat, Clash, and Google's app.

I see. For approving these changes, may I ask how do you determine that these domains are for analytics purposes?

data/category-analytics
```diff
@@ -0,0 +1,5 @@
+# List of all domains being used in log collection or monitor.
+
+analytics.map.qq.com
+app-measurement.com
+in.appcenter.ms
```
KevinZonda, line 5: Would you mind providing this domain's purpose? I found it should be Microsoft App Center SDK. So it should not be kindly analysis tracker?

KevinZonda (changes requested):

KevinZonda:
app-measurement.com is for firebase. [refer](app-measurement.com)  
analytics.map.qq.com is for Tencent Map /cc @IceCodeNew

Loyalsoldier:
Due to the lack of manpower, I personally recommend not collecting or adding "advertising" or "analytics" domains to the project. Instead we'd better use mature third-party projects that are compatible with MIT license.

Or let users generate their own geosite files after we refactor the project. I'd like to do it when I'm free lately.

IceCodeNew:
> Due to the lack of manpower, I personally recommend not collecting or adding "advertising" or "analytics" domains to the project. Instead we'd better use mature third-party projects that are compatible with MIT license.
> 
> Or let users generate their own geosite files after we refactor the project. I'd like to do it when I'm free lately.

I totally agree with that.

KevinZonda:
> Due to the lack of manpower, I personally recommend not collecting or adding "advertising" or "analytics" domains to the project. Instead we'd better use mature third-party projects that are compatible with MIT license.
> 
> Or let users generate their own geosite files after we refactor the project. I'd like to do it when I'm free lately.

It could be a better choice, totally agree with this.

So, should we close this issue? /cc @IceCodeNew @Loyalsoldier

PR #624, opened 2021-09-23
Domains to be added to current lists

molink36:


data/alibaba
```diff
@@ -34,6 +34,7 @@ alibabagroup.com
 alibabaplanet.com
 alibabaued.com
 alibabausercontent.com
+alicdn.com
```
EpLiar, line 37: This domain exists in `aliyun`.

data/rarbg
```diff
@@ -12,3 +12,5 @@ rarbgmirror.org
 rarbgproxy.org
 rarbgprx.org
 rarbgunblock.com
+rarbgway.org
+rarbgaccessed.org
```
EpLiar, line 16: Please mide the alphabetical order.

data/alibaba
molink36: You are right. This domain already exists in aliyun.

data/rarbg
molink36: Done.

EpLiar (approved):

PR #626, opened 2021-09-23
Create cygames

jwstaceyOvO:
popular mobile games from cygames

### DIFF
--- a/data/category-games
+++ b/data/category-games
@@ -1,5 +1,6 @@
 include:blizzard
 include:curseforge
+include:cygames
 include:ea
 include:epicgames
 include:garena
--- a/data/cygames
+++ b/data/cygames
@@ -1,4 +1,6 @@
 cygames.jp
+cygames.co.jp
 worldflipper.jp
+
 worldflipper.akamaized.net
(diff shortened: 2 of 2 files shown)


Loyalsoldier:
I don't like the idea of matching domains by using keywords. Maybe you should use more explicit ones?

jwstaceyOvO:
> I don't like the idea of matching domains by using keywords. Maybe you should use more explicit ones?

Recently, I find that Pretty Derby has changed data link from https://prd-storage-umamusume.akamaized.net/ to https://prd-storage-game-umamusume.akamaized.net/, World Flipper has also added a datalink https://worldflipper.akamaized.net/, which did not exist in previous versions. So, I use keyword to position such links, you don't know when cygames will change them.

jwstaceyOvO:
@Loyalsoldier  Is cygames.co.jp really necessary? I don't find any link of related game api or game data, and such domains don't need special ip to access.

Loyalsoldier:
@jwstaceyOvO  Actually I miss `umamusume.jp`. Added because it belongs to cygames.

PR #628, opened 2021-09-23
Create piratebay

molink36:
The Pirate Bay website and the related domains are added.

### DIFF
--- a/data/piratebay
+++ b/data/piratebay
@@ -1,0 +1,11 @@
+# Piratebay forum
+pirates-forum.org
+
+# Piratebay image hosting site
+torrindex.net
+
+# Piratebay official site
+thepiratebay.org
+
+# Piratebay related site
+apibay.org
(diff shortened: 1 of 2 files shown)


data/piratebay
```diff
@@ -0,0 +1,11 @@
+# Piratebay forum
+pirates-forum.org
+
+# Piratebay image hosting site
+torrindex.net
```
IceCodeNew, line 5: Does this image hosting site only serve pirate bay?

molink36: I believe that `torrindex.net` only serves `thepiratebay.org`, but it might be not limited only to images, but it might contain other static contents, too. It is mentioned in [torrentfreak](https://torrentfreak.com/the-pirate-bays-main-domain-returns-after-month-of-downtime-200411/).

IceCodeNew: Then you may move this domain into another category file.

molink36: Since `torrindex.net` is exclusively used by `thepiratebay.org`, I think that these two websites should be under the same category. Also, my ISP has blocked these domains together to avoid access to `thepiratebay.org` and its mirrors.
What separate category would you suggest for `torrindex.net`?

IceCodeNew: IDK, let's hear from other contributors.

PR #635, opened 2021-09-25
Moving blogger.com to blogspot list

molink36:
`blogger.com` was moved from `google` list to `blogspot` list, which is itself a subcategory of `google`.

### DIFF
--- a/data/blogspot
+++ b/data/blogspot
@@ -1,3 +1,7 @@
+# Main domain which weblogs are registered in
+blogger.com
+
+# Subdomains which weblogs can be accessed from
 blogspot.ae
 blogspot.al
 blogspot.am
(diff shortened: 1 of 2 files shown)


data/google
```diff
@@ -271,7 +271,6 @@ bdn.dev
 beatthatquote.com
 blink.org
 blogblog.com
```
Loyalsoldier, line 273:
```suggestion
```

data/blogspot
```diff
@@ -1,3 +1,7 @@
+# Main domain which weblogs are registered in
+blogger.com
```
Loyalsoldier, line 2:
```suggestion
blogger.com
blogblog.com
```

PR #647, opened 2021-09-28
Update digicert

francis-zhao:
Add `domain: digicert.cn @cn` used in certificate validation in China to `digicert`.

### DIFF
--- a/data/digicert
+++ b/data/digicert
@@ -1,3 +1,4 @@
+digicert.cn @cn
 digicert.com
 digitalcertvalidation.com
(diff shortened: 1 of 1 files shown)


IceCodeNew:
`digicert.cn` should have been covering by category `tld-cn` as its tld is `cn.`
The only possible problem here I could figure out is that the user may have set some proxy rules right above the bypass rules, cause traffic to `digicert.cn` going through proxy in some circumstances.
But that seems to have nothing to do here as a good solution.

PR #652, opened 2021-10-05
add pixnet.in into pixnet

marcushsu:


data/ptt
```diff
@@ -1,3 +1,4 @@
 ptt.cc
 ptt2.cc
 ptt3.cc
+pttweb.cc
```
IceCodeNew, line 4: drop this line as it is duplicated to #653

IceCodeNew (changes requested):
drop the line `pttweb.cc` as it is duplicated to #653

PR #653, opened 2021-10-05
add pttweb.cc into ptt

marcushsu:


data/ptt
```diff
@@ -1,3 +1,4 @@
 ptt.cc
 ptt2.cc
 ptt3.cc
+pttweb.cc
```
IceCodeNew, line 4: This is not the official site.

IceCodeNew (changes requested):
`pttweb.cc` is not the official site.

PR #655, opened 2021-10-09
Create scenesource

molink36:


molink36:
Initially, I was going to put it under `entertainment` category, but then I noticed that, in addition to `movies`, `games`, and `music`, `scenesource` also includes `applications`. Considering this fact, what category would you suggest `scenesource` belongs to?

IceCodeNew:
category-!cn
（（

molink36:
`geolocation-!cn` category has multiple sub-categories. Considering the fact that `scenesource` consists of movies, games, music, programs, and eBooks, which subcategory would you suggest?

PR #684, opened 2021-11-03
Update tencent-ads

Janet-Baker:
Data from capture results on iOS platform.


IceCodeNew:
May you share the capture result?
I could not determine whether your change will impact other normal requests as you are blocking them by a `domain` match.

PR #693, opened 2021-11-15
creat authy and insert into geolocation-!cn

marcushsu:


data/authy
```diff
@@ -0,0 +1 @@
+"authy.com"
```
IceCodeNew, line 1: Clash syntax does not apply here. You will need to change this line.

IceCodeNew (changes requested):

IceCodeNew (approved):

PR #695, opened 2021-11-16
Add Some Disney Domains

yaott2020-666:
From：https://github.com/ACL4SSR/ACL4SSR/blob/master/Clash/Ruleset/DisneyPlus.list


data/disney
```diff
@@ -121,3 +121,18 @@ streamingdisney.net
 thestationbymaker.com
 thisispolaris.com
 watchdisneyfe.com
+adobedtm.com
```
IceCodeNew, line 124: Please sort this list.

IceCodeNew (changes requested):

IceCodeNew (commented):
Request other reviewers based on recently editors.

EpLiar:
I think I can't review these domains. Waiting for other reviews.

data/disney
```diff
@@ -102,8 +115,10 @@ disneyturkiye.com.tr
 disneytvajobs.com
 disneyworld-go.com
 dssott.com
+execute-api.us-east-1.amazonaws.com
```
AkinoKaede, line 118: Hmm, I think this domain of Amazon AWS may be used by other services.

yaott2020-666: Maybe it's a Disney private domain.

These rules are all from: https://github.com/ACL4SSR/ACL4SSR/blob/master/Clash/Ruleset/DisneyPlus.list

IceCodeNew: > Maybe it's a Disney private domain.

I don't think so. Can you find some proof to support this?

yaott2020-666: I'm just guessing. It seems be used by Amazon API: https://docs.aws.amazon.com/general/latest/gr/apigateway.html

I don't have Disney account and could not check it. We need someone can test it.

IceCodeNew: > We need someone can test it.

You can leave that domain behind to get this PR merged sooner.
That domain, once verified, could be appended to this category by submitting another PR.

> (话说我们还有必要全英交流吗)
I think it is necessary. The conversation here could be referred to in another PR, thus it should be "compatible" for almost every contributor.

yaott2020-666: I delete it. If anyone finds it necessary, please open another PR.

PR #697, opened 2021-11-18
add alipaydns and liulishuo

XGFan:


data/geolocation-cn
```diff
@@ -820,11 +820,15 @@ linkvans.com
 linsn.com
 linuxidc.com
 liulanqi.net
+liulishuo.com
```
IceCodeNew, line 823: I would prefer to put these domains together, like this:
```
# 流利说
liulishuo.com
llsapp.com
llscdn.com
llsserver.com

...(other domains that previously exist in this category)
```

IceCodeNew (changes requested):

data/geolocation-cn
XGFan: done

IceCodeNew (approved):

IceCodeNew:
Thanks for your contribution!

PR #701, opened 2021-11-23
Update category-pt

cubercsl:
byr no longer uses the original domain`bt.byr.cn`.


IceCodeNew (approved):

IceCodeNew:
Thanks for your contribution!

PR #703, opened 2021-11-25
add `muwai.com`

alicewww:
dmzj is changing its domain for mobile app

### DIFF
--- a/data/dmzj
+++ b/data/dmzj
@@ -1,2 +1,1 @@
 dmzj.com
-muwai.com
--- a/data/dmzj
+++ b/data/dmzj
@@ -1,1 +1,2 @@
 dmzj.com
+muwai.com
--- a/data/geolocation-cn
+++ b/data/geolocation-cn
@@ -72,6 +72,7 @@ include:le
 include:lizhi
 include:miaopai
 include:mihoyo
+include:muwai
 include:pptv
 include:qingtingfm
 include:sfacg
--- a/data/muwai
+++ b/data/muwai
@@ -1,0 +1,1 @@
+muwai.com


IceCodeNew:
Please provide more info so I can be sure this site is not fraud but "official".

alicewww:
This domain is from [this IOS app](https://apps.apple.com/cn/app/%E5%8A%A8%E6%BC%AB%E4%B9%8B%E5%AE%B6%E7%A4%BE%E5%8C%BA/id1512392614?l=en).

Afaik this is the second time they had changed their domain behind the mobile app (Last time it was dmzj1.com, but not using anymore). I guess this was due to the penalty they received from law enforcements for content issues and they are changing it constantly to avoid being charged again.

My goal is to add this domain to `geolocations-cn` so that my auto-build rules could include it while obeying the current file structure as fine as possible. If you think this proof wasn't good enough, I would create a separate file for it.

IceCodeNew:
> This domain is from [this IOS app](https://apps.apple.com/cn/app/%E5%8A%A8%E6%BC%AB%E4%B9%8B%E5%AE%B6%E7%A4%BE%E5%8C%BA/id1512392614?l=en).
> 
> My goal is to add this domain to `geolocations-cn` so that my auto-build rules could include it while obeying the current file structure as fine as possible. If you think this proof wasn't good enough, I would create a separate file for it.

This iOS app may be a fraud. Have you dug further to determine the relationship between the original person/team and the developer of this application?

alicewww:
Well... Here's the new files :)

IceCodeNew (approved):
Well, I personally think we do not need to concern that the domains could be a fraud or phishing site, as long as they do not confound with other categories.
I'm OK with this PR. And if no one stands up against this PR BEFORE NEXT DAY, I will merge this PR.

PR #711, opened 2021-11-30
Add cloudinary which is media hosting provider

hebo6:


data/cloudinary
```diff
@@ -0,0 +1,2 @@
+cloudinary.com
+cloudinary.net
```
IceCodeNew, line 2: This root domain itself could not be resolved, can you provide any sub-domain in-used of `cloudinary.net` for verifying?

hebo6: cldmo.mo.cloudinary.net

See also: https://who.is/whois/cloudinary.net

IceCodeNew (approved):

PR #732, opened 2021-12-30
Add New Domain threatbook.cn

sinlee1:


data/threatbook
```diff
@@ -0,0 +1,3 @@
+threatbook.cn
+x.threatbook.cn
+s.threatbook.cn
```
IceCodeNew, line 3: The last two domains are covered by the first domain, please delete those two subdomains.

IceCodeNew (changes requested):

IceCodeNew:
BTW, I can't see why we make a category for JUST one domain, especially when that domain has a `.cn` TLD.

database64128:
> BTW, I can't see why we make a category for JUST one domain, especially when that domain has a `.cn` TLD.

I don't use this particular domain, but being able to use `geosite:xxx` instead of having to spell out the full domain can be quite convenient.

IceCodeNew:
> > BTW, I can't see why we make a category for JUST one domain, especially when that domain has a `.cn` TLD.
> 
> I don't use this particular domain, but being able to use `geosite:xxx` instead of having to spell out the full domain can be quite convenient.

That's exactly what I am asking. If you are trying to whitelist this domain, [geolocation-cn](https://github.com/v2fly/domain-list-community/blob/master/data/cn) should have covered it.
If you are including this domain for some different use-case, then I must know that before I can approve this PR.

sinlee1:
Modified

data/threatbook
```diff
@@ -0,0 +1 @@
+threatbook.cn
```
IceCodeNew, line 1: If you are trying to whitelist this domain, geolocation-cn should have covered it.
If you are including this domain for some different use-case, then I must know that before I can approve this PR.

IceCodeNew (changes requested):

data/threatbook
sinlee1: I don't see where the geolocation-cn file is located. If I just want to customize the domain access how do I set it up?

IceCodeNew: > I just checked geolocation-cn and found that there is no such domain inside, so I want to add it

`tld-cn` will cover all `.cn` domains.

sinlee1: Understood, the problem is solved. How do I add a domain name for my personal needs website? If I encounter a domain that does not exist inside geolocation-cn, is it possible to pr submit it again?

IceCodeNew: > How do I add a domain name for my personal needs website?

You can generate your own geosite.dat file. https://github.com/Loyalsoldier/v2ray-rules-dat is a good place to start.

---

> If I encounter a domain that does not exist inside `geolocation-cn`, is it possible to pr submit it again?

Sure. Just issue another PR and wait for someone to review it ;-)

PR #733, opened 2022-01-02
Add Mircosoft NCSI domains

Anshi84:
Network Connection Status Indicator domains:
msftncsi.com
msftconnecttest.com


database64128:
These are already in `private`: https://github.com/v2fly/domain-list-community/blob/master/data/private#L30-L32

PR #734, opened 2022-01-04
Add network security related domains

sinlee1:


IceCodeNew (changes requested):
1. use `category-network-security` as filename.
2. add `@cn` attr or include the whole file in the `geolocate-cn` category. (In that case, the filename should tell it ONLY contains Chinese sites)

data/network_security
```diff
@@ -0,0 +1,4 @@
+fofa.so
+t00ls.cc
+t00ls.net
+tools.team
```
IceCodeNew, line 4: BTW, can you provide the accessible sites' URLs under this domain?

sinlee1: 1. www.fofa.so
2. www.t00ls.cc
3. www.t00ls.net

Both are official websites

sinlee1 (commented):
Changed

data/network_security
IceCodeNew: > * [www.fofa.so](http://www.fofa.so)
> * [www.t00ls.cc](http://www.t00ls.cc)
> * [www.t00ls.net](http://www.t00ls.net)
> 
> Both are official websites

I mean `*.tools.team`

sinlee1: - About fofa.so

`classic.fofa.so`

`faweb.fofa.so`

- About `t00ls.cc` `t00ls.net` `tools.team` 

These three domains are all the same official site, just with different domain names

These 3 domains do not have subdomains, only www domains

IceCodeNew: image

sinlee1: - Situation description

No need to add www, you can directly T00ls.team for access.

The domain name will be redirected to t00ls.cc after 302 hops

Also attached is the official announcement

image

IceCodeNew (approved):

IceCodeNew:
Thanks for your contribution!

PR #739, opened 2022-01-13
Add piwheels.org to python

henry-y-c:


data/python
```diff
@@ -4,3 +4,4 @@ pypi.io
 pypi.org
 python.org
 pythonhosted.org
+piwheels.org
```
IceCodeNew, line 7: > piwheels is a community project by Ben Nuttall and Dave Jones.

I think this domain should be included in another category, or at least add a comment to separate the official projects and the community projects.

IceCodeNew (changes requested):

data/python
henry-y-c: what do you think we add a new file, maybe named as `python-community`?

IceCodeNew: > what do you think we add a new file, maybe named as `python-community`?

I'm OK with that ;-)

henry-y-c: Done

IceCodeNew (approved):

PR #750, opened 2022-01-21
Update category-porn

rootmelo92118:


malikshi:
@rootmelo92118  Sir can you add porn.com to the list?

rootmelo92118:
> @rootmelo92118 Sir can you add porn.com to the list?

ok, but I think you can make your own pull request.

PR #756, opened 2022-01-23
Update youtube

rootmelo92118:


data/youtube
```diff
@@ -171,4 +176,5 @@ youtubei.googleapis.com
 youtubekids.com
 youtubemobilesupport.com
 yt.be
+yt3.ggpht.com
```
IceCodeNew, line 179: Duplicated with https://github.com/v2fly/domain-list-community/pull/756/files#diff-f8917947d128b748ff2ecf053d69193b84de69d30fd0778ca95562ed70f03ea0R5

data/youtube
```diff
@@ -4,9 +4,12 @@ youtube
 ggpht.cn @cn
 ggpht.com
 googlevideo.com
+gvt1.com
+gvt2.com
```
IceCodeNew, line 8: Will there be domains like `gvt3.com` in use? Would it be better to match those domains by the regex rule?

IceCodeNew (changes requested):

data/youtube
rootmelo92118: ok, i removed it.

data/youtube
rootmelo92118: I tried to use regular expression to match gvt domain, but I don't know it's valid or not. The reason is this is  my first time to use regular.
如果有錯誤，請幫我更正，謝謝。

IceCodeNew: > I tried to use regular expression to match gvt domain, but I don't know it's valid or not. The reason is this is my first time to use that. 如果有錯誤，請幫我更正，謝謝。

image
you can test your regex pattern on https://regex101.com/

IceCodeNew (approved):

IceCodeNew:
This PR contains changes that I am not very confident about. Further review is needed before I pull the trigger.

IceCodeNew:
> My data is according to [netify.ai/resources/applications/youtube](https://www.netify.ai/resources/applications/youtube)

Where does `gvt2.com` come from?
image

Loyalsoldier:
I'm afraid these changes have been included in `google` list

https://github.com/v2fly/domain-list-community/blob/7a24a56f23fc0cdda4abf80e9eb458414d4a52d3/data/google#L390-L397

IceCodeNew:
> Actually these changes have been included in `google` list
> 
> https://github.com/v2fly/domain-list-community/blob/7a24a56f23fc0cdda4abf80e9eb458414d4a52d3/data/google#L390-L397

Different categories sharing some common domains is acceptable to me.
I just need to make sure that not a single domain that does not serve YouTube service would be accidentally brought to this category.
If `gvt1.com` does not serve YouTube contents only, then it should be removed from this category, as category `google` would catch this case.

rootmelo92118:
> > My data is according to [netify.ai/resources/applications/youtube](https://www.netify.ai/resources/applications/youtube)
> 
> Where does `gvt2.com` come from? image
截圖 2022-01-24 下午6 49 50

The gvt2.com came from my v2ray log when I watched YouTube.

IceCodeNew:
> The gvt2.com came from my v2ray log when I watched YouTube. And I search it which belong to YouTube.

I see. Yet I still would like to include `gvt*.com` sites in the `google` category, not here.

PR #795, opened 2022-02-27
Update steam

rootmelo92118:


data/steam
```diff
@@ -3,11 +3,13 @@ s.team
 steam-chat.com
 steamcommunity.com
 steamcontent.com
+steamdeck.com
 steamgames.com
 steampowered.com
 steamstatic.com
 steamusercontent.com
 underlords.com
+valve.net
```
IceCodeNew, line 12: can you provide some info about sites under the `valve.net` domain?

KevinZonda, line 12: I double checked this domain, and only find one with ssl subdomain which will guide to a login form. The SSL certification shows it has some relationship with valve corp, however this is not kind of strong evidence I belive.

KevinZonda (changes requested):

data/steam
rootmelo92118: Sorry, this domain was expired. I will remove that immediately.

rootmelo92118:
Sorry, this domain only uses mail service. I will remove that immediately.

IceCodeNew (approved):

IceCodeNew:
Thanks for your contribution!

Issue #830, opened 2022-04-03
Add several domains to the list


jleaves:
Please consider adding the following domains:

Add to "adobe":
- scene7.com (Adobe Dynamic Media Delivery)

Create "goodreads" and add:
- goodreads.com
- gr-assets.com

Add to "linkedin":
- lynda.com (Linkedin Learning)
- linkedin.cn (with tag `@cn`)
- licdn.cn (with tag `@cn`)

Create "apkcombo" and add:
- apkcombo.com

(Ideally, maybe creating a new category "apk-download" about all APK download sites, which includes exsiting "apkpure" and "apkmirror")

IceCodeNew:
PR is welcome.

rootmelo92118:
> PR is welcome.

I have a question. Do you think I need to add a category to include apk download station? The reason is it only have three download stations in this repository. If I create a new category, it will violate the policy or not?

PR #837, opened 2022-04-12
Update adobe

rootmelo92118:
#830

### DIFF
--- a/data/adobe
+++ b/data/adobe
@@ -1,3 +1,4 @@
+include:adobe-activation
 include:typekit
 
 10xfotolia.com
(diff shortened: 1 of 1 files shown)


data/adobe
```diff
@@ -1,3 +1,4 @@
+include:adobe-activation
```
IceCodeNew, line 1: Do not include the unofficial sites.

IceCodeNew (changes requested):

data/adobe
IceCodeNew, line 1:
```suggestion
```

PR #917, opened 2022-05-29
Update geolocation-cn

rootmelo92118:
I am not sure it should be included in `#Services & Software`, but I am sure it should be included in `geolocation-cn`.
Please tell me if I am wrong.


IceCodeNew:
the sites included in that category are all located in China, but only for now.
I do not think it is a good idea to put that category under the `geolocation-cn`

PR #927, opened 2022-06-04
Update category-porn

rootmelo92118:


data/category-porn
```diff
@@ -996,6 +999,7 @@ karupsha.com
 karupspc.com
 katestube.com
 keezmovies.com
+kemono.party
```
IceCodeNew, line 1002: I think this site does not fit the porn category very well.
image

IceCodeNew (changes requested):

data/category-porn
rootmelo92118: > I think this site does not fit the porn category very well. image

Although the front page represent this website is a public archiver, you can see https://kemono.party/posts .

This website storage a lot of pornography content. Actually, this site mainly collects pornography content from the list which is written in front page.

IceCodeNew: u r making a point.

IceCodeNew (approved):

PR #989, opened 2022-07-08
Del javtorrent.re

chika0801:


rootmelo92118:
Sorry, that domain was added by me. I correct its suffix from `re` to `me` now.

rootmelo92118 (approved):

PR #1033, opened 2022-07-25
Update gog

Heptazhou:


Heptazhou:
Known host:
> gog-cdn-lumen-cp77.secure2.footprint.net:443

rootmelo92118 (approved):

PR #1056, opened 2022-08-04
Create tld-opennic

rootmelo92118:
Source : https://wiki.opennic.org/opennic/dot?redirect=1
https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains#Non-IANA_domains

I think the user can use DNS config in v2ray to use specific DNS to resolve the domain which bases on these TLDs more simply.

But, I am not sure whether it should be included in `tld-!cn`. Although it fits the definition of `tld-!cn`, it doesn't make sense to do so because the normal DNS cannot resolve the domains which base on these TLDs.


IceCodeNew (approved):

IceCodeNew:
> But, I am not sure whether it should be included in `tld-!cn`. Although it fits the definition of `tld-!cn`, it doesn't make sense to do so because the normal DNS cannot resolve the domains which base on these TLDs.

I would suggest not to include this file in the `tld-!cn` category, or it will break some existing configurations.

PR #1118, opened 2022-08-23
Update test-pr.yml

rootmelo92118:
If this workflow is just to check the pull request whether is valid or not, my opinion is we don't need to generate some text files with specific categories.


IceCodeNew:
也不太影响测试时间，还是留着吧。未来如果要调整构建功能时，至少可以测试会不会引入 breaking change。

PR #1157, opened 2022-09-09
Update and rename facebook to meta

rootmelo92118:
@IceCodeNew I create this PR because it follows reality. If you don't agree to this PR, you can close directly. You don't need to tell me the reason.


IceCodeNew:
> @IceCodeNew I create this PR because it follows reality. If you don't agree to this PR, you can close directly. You don't need to tell me the reason.

I'm fine with this PR

PR #1159, opened 2022-09-10
Create category-ipfs

rootmelo92118:
source : https://ipfs.github.io/public-gateway-checker/


data/category-ipfs
```diff
@@ -0,0 +1,20 @@
+4everland.io
```
Loyalsoldier, line 1:
```suggestion
Reference: https://ipfs.github.io/public-gateway-checker/

4everland.io
```

PR #1160, opened 2022-09-10
Create tld-tor

rootmelo92118:
This category can make users route tor traffic more conveniently.


Loyalsoldier:
There has been a list called `tor`

PR #1171, opened 2022-09-13
Update dcard

rootmelo92118:


Loyalsoldier:
Cannot verify this. It opens Slack. Source?

rootmelo92118:
> Cannot verify this. It opens Slack. Source?

That domain is using on short link.
https://www.facebook.com/Dcardtopics/
This page is dcard's Facebook fan page.
截圖 2022-09-16 上午7 24 17

PR #1173, opened 2022-09-14
Update twitter & qcloud

KukiSa:
https://help.twitter.com/en references the resource of `cms-twdigitalassets.com`, which is registered through MarkMonitor and can be identified as a Twitter domain.

Please also provide comments on categorizing `twittercommunity.com` and `twitterflightschool.com`.

Replaced enumerated 42 domains using regular expressions:
`tdnsv[1-5]\.(com|net)` 深圳市腾讯计算机系统有限公司 粤B2-20090059-386
`tdnsv([6-9]|10)\.(com|net)` 深圳市腾讯计算机系统有限公司 粤B2-20090059-406
`tdnsv1[1-5]\.(com|net)` 深圳市腾讯计算机系统有限公司 粤B2-20090059-407

### DIFF
--- a/data/qcloud
+++ b/data/qcloud
@@ -29,18 +29,6 @@ cloudsite.vip
 cncqcloud.com
 computeinit.com
 coolsite.vip
-dnse0.cn
-dnse0.com
-dnse1.cn
-dnse1.com
-dnse2.cn
-dnse2.com
-dnse3.cn
-dnse3.com
-dnse4.cn
-dnse4.com
-dnse5.cn
-dnse5.com
 dnsv1.com
 dnsv1.com.cn
 dothework.cn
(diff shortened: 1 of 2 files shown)


rootmelo92118:
If you change the domains to a regular expression, you don't need to give the source again.

KukiSa:
> My opinion of `twittercommunity.com` and `twitterflightschool.com` is they should be included in this category. The reason is it belongs to `twitter`. This definition isn't changed. For example, the domain `pornhubapparel.com` doesn't fit the definition of a pornographic website, but it still belongs to `pornhub`.

The new commit has implemented the changes.


> By the way, you don't need to give the source again if you change the domains to a regular expression.

Regular expressions cover several more domains than the original direct enumeration.

rootmelo92118 (approved):

PR #1179, opened 2022-09-16
Update netlify

KukiSa:


data/netlify
```diff
@@ -1,2 +1,6 @@
-netlify.com
 bitballoon.com
+netlify.app
+netlify.com
+netlifystatus.com
+
+full:d33wubrfki0l68.cloudfront.net
```
rootmelo92118, line 6: Do you have evidence to proof this domain is belonging this category?

KukiSa: The pages of `netlify.com` reference resource from `d33wubrfki0l68.cloudfront.net`.
All other domains can be accessed directly and intuitively concluded.

rootmelo92118 (approved):

data/netlify
IceCodeNew: > The pages of `netlify.com` reference resource from `d33wubrfki0l68.cloudfront.net`. All other domains can be accessed directly and intuitively concluded.

that is not a very strong evidence, could be some third-party sources hosted on that site.

KukiSa: > > The pages of `netlify.com` reference resource from `d33wubrfki0l68.cloudfront.net`. All other domains can be accessed directly and intuitively concluded.
> 
> that is not a very strong evidence, could be some third-party sources hosted on that site.
> 
> @rootmelo92118 @KukiSa

`d33wubrfki0l68.cloudfront.net` is used to host assets when building static web pages.
ref: https://answers.netlify.com/t/complete-urls-for-react-js-bundles/5173

This is a feature provided by Netlify and reflected in the website it builds as shown in the picture below:

IceCodeNew: Good catch!

PR #1182, opened 2022-09-19
Update tld-!cn

rootmelo92118:


data/tld-!cn
```diff
@@ -468,6 +470,15 @@ teva # Teva Pharmaceutical Industries Limited
 toray # Toray Industries, Inc.
 weather # The Weather Channel LLC
 
+# Sex & Porn
+adult # Adult content
+gay # Homosexuality
+lgbt # Lesbian, Gay, Bisexual, Transgender
```
IceCodeNew, line 476: might not all the sites under `.lgbt` are about porn.
I would suggest we put that TLD under no categories, until we have built up good knowledge about it.

IceCodeNew (changes requested):

data/tld-!cn
rootmelo92118: > might not all the sites under `.lgbt` are about porn. I would suggest we put that TLD under no categories, until we have built up good knowledge about it.

Ok, I think it can be move to `Miscellaneous Domains`.

IceCodeNew: > > might not all the sites under `.lgbt` are about porn. I would suggest we put that TLD under no categories, until we have built up good knowledge about it.
> 
> Ok, I think it can be moved to `Miscellaneous Domains`.

That would be a good start. Mark this conversation as resolved when you finished fixing that.

data/tld-!cn
```diff
@@ -468,6 +471,14 @@ teva # Teva Pharmaceutical Industries Limited
 toray # Toray Industries, Inc.
 weather # The Weather Channel LLC
 
+# Sex & Porn
+adult # Adult content
+gay # Homosexuality
```
IceCodeNew, line 476: This one should be moved out too.

IceCodeNew (changes requested):

data/tld-!cn
rootmelo92118: Although I can move them to `Miscellaneous Domains`, the definition of sex here not only means pornography but also means sexuality.

IceCodeNew (approved):

data/tld-!cn
IceCodeNew: > Although I can move them to `Miscellaneous Domains`, the definition of sex here not only means pornography but also means sexuality.

I see. But I still get some strange felling about it.

PR #1227, opened 2022-09-30
Add qcc to geolocation-cn to fix qcc.com access issue

yukeiyang:
qcc.com is one popular enterprise inquiry website in China. The domain qcc.com was changed from original domain qichacha.com.

There are 4 subdomains of qichacha.com that still be depended, there is one issue to access,  After add it, the issue will gone.


rootmelo92118:
The `qcc` is already included by `category-enterprise-query-platform-cn` and that is already included in `geolocation-cn`.

PR #1230, opened 2022-10-01
google: sync `@cn` rules with upstream

database64128:


data/google
```diff
@@ -530,61 +530,58 @@ zukunftswerkstatt.de
 
 # Domains and services below are available in China mainland
 # Use in config file like this: "geosite:google@cn"
-full:csi-china.l.google.com @cn
-full:gstaticadssl.l.google.com @cn
```
IceCodeNew, line 534: 这两个为什么删掉了啊？

IceCodeNew: @database64128

database64128: I tested and they did not appear to have access points in China.

PR #1235, opened 2022-10-04
update Hulu

WordsWorthLess:
```
hulu.playback.edge.bamgrid.com
```
The domain above is for geolocation check, leaving it unproxied would cause playback issues with error codes 'p-dev3xxx'

And this conflicts with the entry "bamgrid.com" of the disney ruleset, make sure the hulu ruleset is prior to the disney one when  using different proxies for these two streaming service

### DIFF
--- a/data/hulu
+++ b/data/hulu
@@ -45,3 +45,4 @@ payhulu.com
 registerhulu.com
 thehulubraintrust.com
 wwwhuluplus.com
+full:hulu.playback.edge.bamgrid.com


rootmelo92118 (approved):

data/hulu
```diff
@@ -45,3 +45,4 @@ payhulu.com
 registerhulu.com
 thehulubraintrust.com
 wwwhuluplus.com
+full:hulu.playback.edge.bamgrid.com
```
IceCodeNew, line 48: 下次可以在 full 类的条目和 domain 类的条目之间加一个空格
@rootmelo92118 @WordsWorthLess

PR #1238, opened 2022-10-08
apple: sync @cn rules with upstream

Loyalsoldier:
Remove `stocks-sparkline.apple.com` and `weather-data.apple.com` due to duplication.

### DIFF
--- a/data/apple
+++ b/data/apple
@@ -798,7 +798,6 @@ full:app-site-association.cdn-apple.com @cn
 full:appldnld.apple.com @cn
 full:appldnld.g.aaplimg.com @cn
 full:appleid.cdn-apple.com @cn
-full:apps.apple.com @cn
 full:apps.mzstatic.com @cn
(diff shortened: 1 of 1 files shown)


data/apple
```diff
@@ -796,10 +794,12 @@ full:a4.mzstatic.com @cn
 full:a5.mzstatic.com @cn
 full:adcdownload.apple.com.akadns.net @cn
 full:adcdownload.apple.com @cn
+full:app-site-association.cdn-apple.com @cn
 full:appldnld.apple.com @cn
 full:appldnld.g.aaplimg.com @cn
+full:appleid.cdn-apple.com @cn
+full:apps.apple.com @cn
```
database64128, line 801: See #503. Maybe comment this line out?

Loyalsoldier, line 801:
```suggestion
```

Loyalsoldier: Fixed.

PR #1286, opened 2022-10-28
add onion site for `nytimes`

zydou:
https://open.nytimes.com/https-open-nytimes-com-the-new-york-times-as-a-tor-onion-service-e0d0b67b7482


rootmelo92118:
I don't think it should be included here because it cannot be routed correctly.

rootmelo92118:
You can refer to the category of `tor`.

Issue #1306, opened 2022-11-08
add category-tracker


Hubedge:
I'm aware of https://github.com/v2fly/domain-list-community/blob/master/data/category-pt but we lack a category for trackers.

https://github.com/ngosang/trackerslist would be a good starting point.

IceCodeNew:
Naming the category as "category-tracking" could lead to misunderstanding.  
How about `category-public-tracker`?  
Also, PR is welcomed.  
  
**Please be aware that copying & pasting the third-party maintained list is not encouraged, doing so would bring the risk of syntax error and violating their license.**

Hubedge:
`category-public-tracker` sounds good.

I will try to put together a PR from scratch.

PR #1309, opened 2022-11-09
Add `category-public-tracker`

Hubedge:
#1306

Not an exhaustive list. Just some trackers pulled from a few torrents downloaded from popular torrent site.


IceCodeNew:
this list needs to be sorted.

IceCodeNew (approved):

PR #1326, opened 2022-11-21
Update and rename facebook to meta

wangling12:


rootmelo92118:
The category `meta` doesn’t fit the definition of social media.

PR #1340, opened 2022-12-02
Create category-ir-blocked-list

rootmelo92118:
Fix #1337 

But I don't think this is a good way to solve the problem.

PR #1344, opened 2022-12-06
Add iranian ITO list

debMan:
There are some points should I mention here:

- Not all of these domains are Iran-only access, many of them are available both inside and outside.
- Not all Iranian domains are listed here. This list just contains domains that are listed on [Information Technology Organization of Iran](https://eservices.ito.gov.ir/Page/IPList).
- The list excluded IPv4 addresses listed there.


debMan:
Will fix  #1340
I suggest to change pull request name to `category-ir-ito`.

debMan:
Sure. Thanks

PR #1346, opened 2022-12-06
Added ITO listed iranian domains

IceCodeNew:
Will fix https://github.com/v2fly/domain-list-community/pull/1340 and close https://github.com/v2fly/domain-list-community/issues/1337.

PR #1375, opened 2022-12-25
create IR

mmrabbani:
This is list of Iranian hosted domain. This files contains website that will not be blocked by Iranian government (website hosted inside Iran and all .ir websites, because Iran won't block .ir websites). Users from Iran can exclude this list from their tunnel.


rootmelo92118:
First, I don't think this is a good idea to use regex to match `*.ir` domain names. `ir` is equal to `regex:.+\.ir$`.
Second, I suggest renaming this file to `geolocation-ir`.

rootmelo92118:
What is the difference between #1375 to #1346? If you are the same, we only can merge one.

mmrabbani:
> What is the difference between #1375 to #1346? If you are the same, we only can merge one.

They are totally different. #1375 contains all Iranian domains that use can direct them. #1376 contains Iranian ad websites that user can block.

rootmelo92118:
> > What is the difference between #1375 to #1346? If you are the same, we only can merge one.
> 
> They are totally different. #1375 contains all Iranian domains that use can direct them. #1376 contains Iranian ad websites that user can block.

I said #1346, not #1376.

mmrabbani:
> > > What is the difference between #1375 to #1346? If you are the same, we only can merge one.
> > 
> > 
> > They are totally different. #1375 contains all Iranian domains that use can direct them. #1376 contains Iranian ad websites that user can block.
> 
> I said #1346, not #1376.

They both have a common goal. but they gather list from two separate sources. Most of these two list are common. You can merge one. After that, the merged one can be more complete with help of the other.
If you select #1346 , I have a comment about this. ITO is only one source of Iran hosted domains. maybe choosing more general name for file name would be better.

IceCodeNew:
> > > > What is the difference between #1375 to #1346? If you are the same, we only can merge one.
> > > 
> > > 
> > > They are totally different. #1375 contains all Iranian domains that use can direct them. #1376 contains Iranian ad websites that user can block.
> > 
> > 
> > I said #1346, not #1376.
> 
> They both have a common goal. but they gather list from two separate sources. Most of these two list are common. You can merge one. After that, the merged one can be more complete with help of the other. If you select #1346 , I have a comment about this. ITO is only one source of Iran hosted domains. maybe choosing more general name for file name would be better.

IMO it looks like closing #1346 and letting #1376 get merged is the right call.
#1346 is too big for reviewing. It should be separated into several PR which are much more little, and #1376 would be the first step.

@mmrabbani @rootmelo92118

rootmelo92118:
> > > > > What is the difference between #1375 to #1346? If you are the same, we only can merge one.
> > > > 
> > > > 
> > > > They are totally different. #1375 contains all Iranian domains that use can direct them. #1376 contains Iranian ad websites that user can block.
> > > 
> > > 
> > > I said #1346, not #1376.
> > 
> > 
> > They both have a common goal. but they gather list from two separate sources. Most of these two list are common. You can merge one. After that, the merged one can be more complete with help of the other. If you select #1346 , I have a comment about this. ITO is only one source of Iran hosted domains. maybe choosing more general name for file name would be better.
> 
> IMO it looks like closing #1346 and letting #1376 get merged is the right call. #1346 is too big for reviewing. It should be separated into several PR which are much more little, and #1376 would be the first step.
> 
> @mmrabbani @rootmelo92118

I agree.

IceCodeNew:
> Unfortunately, after the private discussion, we decide to close this pull request directly.

It might be better to put the conversation here. You may use ChatGPT to do the translation if it is available to you.
Just do not forget to send the translated conversation to me for review before you publish it publicly.

@rootmelo92118

PR #1376, opened 2022-12-25
Create ir-ads

mmrabbani:
Adding the list contains all ad services in Iran.
Iranian users can block this list on client or server side.


rootmelo92118 (approved):

IceCodeNew:
let me have a look

data/ir-ads
```diff
@@ -0,0 +1,361 @@
...
+clickaval.com
+clickyab.com
+chabok.io
+
+congoro.com
+congoro.ir
+
+# >> D <<
+24d.ir
+l.24d.ir
```
rootmelo92118, line 64: The domain `24d.ir` will match the domain `*.24d.ir`, so you don't need to add `l.24d.ir`. Please check the domain names submitted by you. And fix this kind of situation.

rootmelo92118 (changes requested):

data/category-ads
```diff
@@ -54,6 +54,7 @@ include:xiaomitv-ads
 include:yahoo-ads
 include:youku-ads
 include:zynga-ads
+include:ir-ads
```
rootmelo92118, line 57: Sort it, please.

rootmelo92118 (changes requested):

mmrabbani:
@rootmelo92118 
Thank you for comments. Both comments are implemented.

data/ir-ads
```diff
@@ -0,0 +1,358 @@
...
+
+scripts-ads.s3.ir-thr-at1.arvanstorage.com
+error-tracking.arvancloud.com
+
+adexo.ir
+adskav.com
+advn.ir
+
+amaroid.net
+lnk.amaroid.net
```
rootmelo92118, line 17: this

data/ir-ads
```diff
@@ -0,0 +1,358 @@
...
+adpulse.ir
+adsima.net
+adtrace.io
+affili.ir
+amarfa.ir
+ayyaar.ir
+# ^ Used by irna.ir
+
+adro.co
+affstat.adro.co
```
rootmelo92118, line 39: this

data/ir-ads
```diff
@@ -0,0 +1,358 @@
...
+behinava.com
+radar.bayan.ir
+backority.ir
+
+# >> C <<
+cayot.ir
+# Example: tamasha.com/v/DDlMB
+
+chavosh.org
+ck.chavosh.org
```
rootmelo92118, line 53: this

rootmelo92118 (changes requested):
Remove three domains because they are duplicated.

rootmelo92118:
@IceCodeNew I think it doesn't have a problem, just need to change the format and sort it. What do you think about that?

rootmelo92118 (approved):

IceCodeNew:
> @IceCodeNew I think it doesn't have a problem, just need to change the format and sort it. What do you think about that?

sorry for missing the notification, I will review this PR later today.

data/ir-ads
```diff
@@ -0,0 +1,355 @@
...
+amaroid.net
+
+analytics.aasaam.com
+analytics-2.aasaam.com
+analytics-3.aasaam.com
+
+adwised.com
+adwisedfs.com
+adwised.azureedge.net
+# TODO: Remove adwised.azureedge.net, but navatv.com and baran-music.com still links to it
```
IceCodeNew, line 25: why not remove this site now? If it will break sites like tv or music service, then we should not use the blocking strategy based on domain.
Extensions like the ublock can do a much better job.

data/ir-ads
```diff
@@ -0,0 +1,355 @@
...
+
+adivery.com
+adnegah.net
+adpulse.ir
+adsima.net
+adtrace.io
+affili.ir
+amarfa.ir
+ayyaar.ir
+# ^ Used by irna.ir
```
IceCodeNew, line 35: What does the phrase `used by` means? Does it mean this site is also necessary for the `irna.ir` to function normally?

data/ir-ads
```diff
@@ -0,0 +1,355 @@
...
+
+# >> E <<
+
+# >> F <<
+facepop.org
+farsbux.ir
+fastclick.ir
+
+farakav.com
+#  TODO: Check for breakages (used by tamasha.com), investigate into more tracking
```
IceCodeNew, line 82: If in doubt, please do not introduce it.

data/ir-ads
```diff
@@ -0,0 +1,355 @@
...
+phonroid.com
+# ^ Example: par30games.net
+
+popina.ir
+
+popunder.ir
+# TODO: Remove it, Used by marzfun.com
+
+popupdl.ir
+#  TODO: Remove it, Used by dlfox.com
```
IceCodeNew, line 150: please do it NOW

IceCodeNew (changes requested):

IceCodeNew:
Thanks for the detailed comments, it really makes the reviewing much easier.
I can not help to express my feelings about seeing someone put that much effort to make a high-quality pull request. You did a great job. 👍

Loyalsoldier:
I notice that this PR is based on the repo https://github.com/MasterKia/PersianBlocker , which is licensed under `AGPL-3.0 license`. And it is NOT compatible with our repo.

IceCodeNew:
> I notice that this PR is based on the repo [MasterKia/PersianBlocker](https://github.com/MasterKia/PersianBlocker) , which is licensed under `AGPL-3.0 license`. And it is NOT compatible with our repo.

Good catch!
@mmrabbani please be noticed.

data/ir-ads
mmrabbani: It means these ad websites are used in irna.ir (Iran largest news agency). Irna is working great after blocking them.

data/ir-ads
mmrabbani: Done.

data/ir-ads
mmrabbani: Removed.

data/ir-ads
mmrabbani: Removed.

data/ir-ads
```diff
@@ -0,0 +1,343 @@
+# List of ads and tracking company in Iran.
+# Thanks to PersianBlocker project
```
IceCodeNew, line 2: giving credit is not enough, you CAN NOT copy content from a project that has a conflict license

IceCodeNew:
> `PersianBlockerHosts` is based the main `PersianBlocker` filterlist and it's used by different DNS blockers; some of them (e.g. PiHole) won't block `ads.example.com` if you only use `example.com`. If I remove `ads.example.com` from the list then those DNS blockers will be left unprotected. V2fly should use deduplication and remove those "duplicate" domains.

That is correct, though we do have code for that kind of duplicating, it is not encouraged for contributors to copy content from other projects.
That is because not all syntax differences can be solved by the duplicating function, and doing so will have the copyright/copyleft problem as well as the sync problem (like you had mentioned about)


> I guess now it's my turn:
> 
> * Manually copying the list's content into your project isn't fruitful; adserver/tracker domains change all the time and the list needs to be constantly updated => you'll need someone to keep the two projects in sync, bad idea.
>   GitHub Actions is a good idea as it can fetch the list daily, also it can remove duplicated entries by itself instead of putting the burden on upstream lists.
>   @bootmortis can help here.

You are making a point, that is the very reason I closed this PR.
As for how to introduce blocklist from other projects without bringing up the sync problem, I would personally recommend the project https://github.com/Loyalsoldier/domain-list-custom
It gives users a good surface to manipulate the content of any category of this project. (See https://github.com/Loyalsoldier/domain-list-custom/blob/cc8590a9489c8f5fa484f80cd32beb1449cb10e8/.github/workflows/build.yml#L42 )
You can simply download the blocklist you want to merge into this project, solve the syntax problem, and then you are good to go!


> Maybe this project should reconsider its licensing choice so that you could benefit from all of those regional lists.
I do not think it is necessary for us to change the license to the one which restricts more of what users can do.
The problem can not be solved here, can be solved in a fork ;-)

@MasterKia @mmrabbani @rootmelo92118

IceCodeNew:
> > license to the one which restricts more of what users can do
> 
> A copyleft license ensures that **all** users will receive the same freedoms that the author has intended. Noone can make some modifications to a copyleft-licensed project and then redistribute it without giving the same freedoms to other users.
> 
> Feel free to use any of these projects: [nimasaj/uBOPa@`master`/uBOPa_Pihole.txt](https://github.com/nimasaj/uBOPa/blob/master/uBOPa_Pihole.txt?rgh-link-date=2023-02-22T05%3A44%3A02Z) (Expat/MIT license)
> 
> [farrokhi/adblock-iran@`master`/pihole.txt](https://github.com/farrokhi/adblock-iran/blob/master/pihole.txt?rgh-link-date=2023-02-22T05%3A44%3A02Z) (2-Clause BSD License)
> 
> Though these lists might contain obsolete/dead domains and are not very responsive to contributions.

Nah, I did not choose words correctly, let me clarify it. By _restricts more of what users can do_, I meant to say there are _more rules that users have to conform_. 
  
I am afraid the decision is final. Whatever the license is, switching from MIT to it is a cost to me (and also to the users). The benefit of importing a third-party blocklist is just not enough to compensate for that cost, not to mention we have no plan to support importing a third-party blocklist.
  
As I said before, the project https://github.com/Loyalsoldier/domain-list-custom is easy to custom to import other blocklists, and it happens to use this project as its very foundation.
IMO any modification that is not accepted by this project could move to a fork of https://github.com/Loyalsoldier/domain-list-custom

IceCodeNew:
> > As I said before, the project [Loyalsoldier/domain-list-custom](https://github.com/Loyalsoldier/domain-list-custom) is easy to custom to import other blocklists, and it happens to use this project as its very foundation.
> > IMO any modification that is not accepted by this project could move to a fork of [Loyalsoldier/domain-list-custom](https://github.com/Loyalsoldier/domain-list-custom)
> 
> Is there a documentation (preferably in English) about adding a new list to that project? Thanks.
> 
> @Loyalsoldier @IceCodeNew

IIRC merging a third-party domain list is not an intentionally implemented feature, and there is no document showing you how to do this.
You can create a topic at the GitHub Discussions shall you have further questions. Next time you may only notify me to join the discussion, without disturbing Loyalsoldier.

Here is a simple example for abusing https://github.com/Loyalsoldier/domain-list-custom :
```diff
      - name: Append attribute rules
        if: ${{ env.NeedToSync }}
        run: |
-         echo "include:geolocation-!cn @cn" >> ./domain-list-community/data/cn
-         echo "include:geolocation-cn @!cn" >> ./domain-list-community/data/geolocation-\!cn
+         $(type -P curl) -LRq --retry 5 --retry-delay 10 --retry-max-time 60 'https://raw.githubusercontent.com/Loyalsoldier/domain-list-custom/release/cn.txt' -o './domain-list-community/data/cn'
+         $(type -P curl) -LRq --retry 5 --retry-delay 10 --retry-max-time 60 'https://raw.githubusercontent.com/Loyalsoldier/domain-list-custom/release/geolocation-!cn.txt' -o './domain-list-community/data/geolocation-\!cn'
+
+         $(type -P curl) -LRq --retry 5 --retry-delay 10 --retry-max-time 60 'https://raw.githubusercontent.com/farrokhi/adblock-iran/master/pihole.txt' -o './temp_file'
+         # solve the syntax problem if it is necessary
+         cat './domain-list-community/data/category-ads-all' >> './temp_file'
+         mv -f './temp_file' './domain-list-community/data/category-ads-all'
```

Issue #1395, opened 2022-12-31
IPv6地址作为SNI时会匹配private规则的可能性


kabegoe:
当IPv6地址作为SNI时，会匹配
https://github.com/v2fly/domain-list-community/blob/7d24ecd45c7aecc3e23bfc4dce5964dfdc505858/data/private#L121-L122
`private`的这条规则。

巧合的是，IPv4地址作为SNI时并不会出现问题。其原因是IPv4地址本身包含`.`，恰好会被排除掉。

是否可以考虑把此规则替换成`regexp:^[^.:]+$`?

IceCodeNew:
> 当IPv6地址作为SNI时，会匹配
> 
> https://github.com/v2fly/domain-list-community/blob/7d24ecd45c7aecc3e23bfc4dce5964dfdc505858/data/private#L121-L122
> 
> 
> `private`的这条规则。
> 巧合的是，IPv4地址作为SNI时并不会出现问题。其原因是IPv4地址本身包含`.`，恰好会被排除掉。
> 
> 是否可以考虑把此规则替换成`regexp:^[^.:]+$`?

Good catch

database64128:
It's important to note that this issue was originally posted as SagerNet/sing-box#305.

I don't think this project should account for the possibility that some programs may match **string representations of IP addresses** against **domain** rules. Especially in this case, it should be a trivial fix for sing-box. sing-box should have checked whether the SNI is a string representation of an IP before matching it against domain rules.

IceCodeNew:
> It's important to note that this issue was originally posted as SagerNet/sing-box#305.
> 
> I don't think this project should account for the possibility that some programs may match **string representations of IP addresses** against **domain** rules. Especially, in this case, it should be a trivial fix for sing-box. sing-box should have checked whether the SNI is a string representation of an IP before matching it against domain rules.

You got a point here. Yet I still would like to make the regex expression more accurate.

nekohasekai:
As far as I know, this SNI extension is not sent when connecting to an IP address.

mazzz1y:
> I don't think this project should account for the possibility that some programs may match string representations of IP addresses against domain rules. Especially in this case, it should be a trivial fix for sing-box. sing-box should have checked whether the SNI is a string representation of an IP before matching it against domain rules.

> As far as I know, the SNI extension is not sent when connecting to an IP address.

[...]

PR #1400, opened 2023-01-03
fix #1395

IceCodeNew:

PR #1412, opened 2023-01-06
Dotless domain regex fix

mazzz1y:
Dotless domain should match these rules:

1) Characters should only be a-z | A-Z | 0-9 and dash(-)
2) Should not start or end with dash (-) (e.g. -domain-)
3) Should be between 1 and 63 characters long

Currently it can be any string without dots. 

Related issue: https://github.com/v2fly/domain-list-community/issues/1395

### DIFF
--- a/data/private
+++ b/data/private
@@ -119,7 +119,7 @@ b.e.f.ip6.arpa
 127.100.in-addr.arpa
 
 # Dotless domains
-regexp:^[^.]+$
+regexp:^[-0-9A-Za-z]+$
 
 # Aruba router
 full:instant.arubanetworks.com


database64128:
> Dotless domain should match these rules

Can you reference the standard document that defines these rules?

mazzz1y:
> document that defines these rules

https://www.rfc-editor.org/rfc/rfc1035

> ...
> The labels must follow the rules for ARPANET host names.  They must
start with a letter, end with a letter or digit, and have as interior
characters only letters, digits, and hyphen.  There are also some
restrictions on the length.  Labels must be 63 characters or less.
>
> For example, the following strings identify hosts in the Internet:
> A.ISI.EDU XX.LCS.MIT.EDU SRI-NIC.ARPA
> ...

rootmelo92118:
In my opinion, I suggest using a shorter regex to match the dotless domain. Although your syntax is correct, I don't think some parts are necessary. For example, First, I think you don't need to limit the domain to 63 characters because this kind of domain does not exist. Nobody will send the domain which is over 63 characters to v2ray. Second, the domain start or ends with single or multiple dashes does not exist. In other words, your regex is trying to prevent the cases which are impossible.
I think your regex just needs to process the situations which are possible. e.g.:`regexp:^[-0-9A-Za-z]+$`
Of course, it is just my personal opinion, so if you don't want to modify it. That's fine.

data/private
```diff
@@ -119,7 +119,7 @@ b.e.f.ip6.arpa
 127.100.in-addr.arpa
 
 # Dotless domains
-regexp:^[^.]+$
+regexp:^[A-Za-z0-9]([A-Za-z0-9-]{1,61}[A-Za-z0-9]|[A-Za-z0-9]|$)$
```
rootmelo92118, line 122: In my opinion, I suggest using a shorter regex to match the dotless domain. Although your syntax is correct, I don't think some parts are necessary. For example, First, I think you don't need to limit the domain to 63 characters because this kind of domain does not exist. Nobody will send the domain which is over 63 characters to v2ray. Second, the domain start or ends with single or multiple dashes does not exist. In other words, your regex is trying to prevent the cases which are impossible.
I think your regex just needs to process the situations which are possible. e.g.:regexp:^[-0-9A-Za-z]+$
Of course, it is just my personal opinion, so if you don't want to modify it. That's fine.

rootmelo92118 (changes requested):

data/private
rootmelo92118: P.S. If you don't want to modify it, just press "Resolve conversation".

mazzz1y: > Nobody will send the domain which is over 63 characters to v2ray. Second, the domain start or ends with single or multiple dashes does not exist. In other words, your regex is trying to prevent the cases which are impossible.

We can say the same about https://github.com/v2fly/domain-list-community/issues/1395 issue. Nobody will send IP address as SNI header, but errors happens as we can see.

But I agree with `^[-0-9A-Za-z]+$` and I think that it is enough.

Also, I tried to use hostname with trailing dash in linux environment and looks like it works:
```
# echo '127.0.0.1 test-hostname-' >> /etc/hosts
# ping test-hostname-
PING test-hostname- (127.0.0.1) 56(84) bytes of data.
64 bytes from localhost.localdomain (127.0.0.1): icmp_seq=1 ttl=64 time=0.086 ms
```

So not everyone follows these rules.

rootmelo92118: >We can say the same about https://github.com/v2fly/domain-list-community/issues/1395 issue. Nobody will send IP address as SNI header, but errors happens as we can see.

Well, that is impossible in v2ray or xray, but it happens in sing-box, so this problem needs to solve.

And thank you for accepting my suggestion.

rootmelo92118 (approved):

rootmelo92118 (approved):
@IceCodeNew I think this PR can be merged, how about you?

IceCodeNew:
Suppressed by #1400

rootmelo92118:
> > @IceCodeNew I think this PR can be merged, how about you?
> 
> I prefer the previous commit.

Why? I think the new regex pattern can cover more situations of the dotless domain. Just like he said, it is not everyone following RFC1035. And the dotless domain name only uses in the local area network. It will not get any negative effect on WAN. So, I think that is better than the previous one.

IceCodeNew:
> > > @IceCodeNew I think this PR can be merged, how about you?
> > 
> > 
> > I prefer the previous commit.
> 
> Why? I think the new regex pattern can cover more situations of the dotless domain. Just like he said, it is not everyone following RFC1035. And the dotless domain name only uses in the local area network. It will not get any negative effect on WAN. So, I think that is better than the previous one.

In practice, the `private` category is used for matching very few domains, it does not make sense to cover more situations of the dotless domain.
if there are domains not following RFC1035, they should not be matched by a rarely used rule. And users who really want them to be matched can write their own rules.
I do not want to break something. So I prefer to follow the specifications.

PR #1294, opened 2022-11-02
Add category-ir

molaeiali:
Iranian websites that will not open without an Iranian IP

PR #1438, opened 2023-01-27
Create geolocation-ir

rootmelo92118:
This PR combine the content of  #1346 and #1375.


IceCodeNew:
> In my opinion, hand-picking domains, like #1294 is way better than using all the domains listed in `ITO`. It can have many mistakes and cause a lot of problems for the end user.
> 
> Right now, I can see two domains hidden in 12000 domains that aren't related to Iran!
> 
> * `ooklaserver.net` which is for speedtest.net
> * `googleapis.com`, which is for google.com
> 
> There might be more of these examples hidden in all of these domains.

Good catch! Thanks for the advice ;-)

PR #1454, opened 2023-02-01
Create mingyuanyun

clcc2019:


rootmelo92118:
This category should not be created because it has only one domain. Maybe you should add it in `geolocation-cn`.
https://github.com/v2fly/domain-list-community/blob/e8f321e65238414463f47269f3641ee143ae4b0f/data/geolocation-cn#L219

PR #1481, opened 2023-02-18
Update openai

SakuraSakuraSakuraChan:
Update openai related domain

### DIFF
--- a/data/openai
+++ b/data/openai
@@ -1,2 +1,12 @@
 # Main domain
 openai.com
+ai.com
+
+# Api
+auth0.com
+sentry.io
+unpkg.com
+
+# CDN
+full:openaiapi-site.azureedge.net
+full:challenges.cloudflare.com
--- a/data/openai
+++ b/data/openai
@@ -1,6 +1,6 @@
 # Main domain
-openai.com
 ai.com
+openai.com
 
 # CDN
 full:openaiapi-site.azureedge.net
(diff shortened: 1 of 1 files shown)


data/openai
```diff
@@ -1,2 +1,12 @@
...
+ai.com
+
+# Api
+auth0.com
+sentry.io
+unpkg.com
+
+# CDN
+full:openaiapi-site.azureedge.net
+full:challenges.cloudflare.com
```
rootmelo92118, line 12: We need more evidence to prove lines `6` `7` `8` and `12` belong to `openai`.

rootmelo92118 (changes requested):

data/openai
SakuraSakuraSakuraChan: They don't, openai is using these services

rootmelo92118: They should not include in this file if they are not belonging to `openai` or are only used on `openai`.

IceCodeNew:
@rootmelo92118 please notify me when you think this PR is ready to go.

rootmelo92118:
@IceCodeNew Done.

data/openai
```diff
@@ -1,2 +1,6 @@
 # Main domain
 openai.com
+ai.com
```
IceCodeNew, line 3:
```suggestion
ai.com
openai.com
```

IceCodeNew (approved):

PR #1483, opened 2023-02-19
feat: add quay to category-dev

kunish:
As the title suggests, this PR adds `quay` to category-dev.txt

This includes `quay.io` for example to support [quay container image registry](https://quay.io)

### DIFF
--- a/data/category-dev
+++ b/data/category-dev
@@ -123,6 +123,7 @@ php.net
 putty.org
 
 postgresql.org
+quay.io
 r-project.org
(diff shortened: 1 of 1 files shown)


rootmelo92118 (changes requested):
If this category just has 1 domain, you should add the domain into `category-dev` directly. You don't need to add a new category and include it in `category-dev`.

rootmelo92118 (approved):
Thanks for your contribution and collaboration.

PR #1496, opened 2023-02-23
add ciweimao

OpportunityLiu:
https://data.chinaz.com/company/5d513032f1ae7a274b239682

image

### DIFF
--- a/data/ciweimao
+++ b/data/ciweimao
@@ -1,0 +1,2 @@
+ciweimao.com
+kuangxiangit.com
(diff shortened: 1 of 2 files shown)


data/ciweimao
```diff
@@ -0,0 +1,2 @@
+ciweimao.com
+kuangxiangit.com
```
IceCodeNew, line 2: can not tell whether this site is owned by the `ciweimao` from the whois info
resources from this site were loaded when accessing the `ciweimao` is not solid proof.

IceCodeNew (changes requested):

data/ciweimao
OpportunityLiu: image

IceCodeNew (approved):

PR #1503, opened 2023-02-24
Update category-porn

rootmelo92118:


data/category-porn
```diff
@@ -5207,6 +5208,7 @@ regexp:(^|\.)[1-9]+zipai\.(com|net)$
 regexp:(^|\.)[1-9]00caocao\.com$
 regexp:(^|\.)[3-9]0000sao\.com$
 regexp:(^|\.)[a-z][1-9][0-9][a-z]\.com$
+regexp:(^|\.).*fulibl.*\.(cc|com|org|net|xyz)$
```
IceCodeNew, line 5211: it does look right to me, what are the domains you are trying to match against?

IceCodeNew (changes requested):

data/category-porn
rootmelo92118:
```
bfulibl.net
fulibl.cc
fulibl.com
fulibl.net
fulibl.org
fulibl.xyz
ifulibl.net
```

Sometimes, it will add an alphabet before `fulibl`. Actually, I wanna use the syntax `keyword:`, but I am afraid it will match the domain which should not be matched.

data/category-porn
```diff
@@ -2842,7 +2835,6 @@ minisexdoll.com
 minnano-av.com
 minuporno.com
 miohentai.com
-mirar.xxx
```
rootmelo92118, line 2845: Unavailable.

data/category-porn
```diff
@@ -2918,7 +2910,6 @@ mrpornlive.com
 mrsSiren.com
 mrskin.com
 mrvideosdesexo.xxx
-mt-ssul1.com
```
rootmelo92118, line 2921: Unavailable.

data/category-porn
```diff
@@ -3374,7 +3365,6 @@ pornacho.com
 pornadoo.com
 pornaf.com
 pornaffected.com
-pornagent.xyz
```
rootmelo92118, line 3377: Unavailable.

data/category-porn
```diff
@@ -4813,7 +4798,6 @@ xb18.me
 xbabe.com
 xbahis33.com
 xbahis44.com
-xbahis55.com
```
rootmelo92118, line 4816: Unavailable.

data/category-porn
```diff
@@ -4951,7 +4935,6 @@ xvidzz.com
 xvirtual.com
 xvxx.stream
 xvxxtube.com
-xwebporn.com
```
rootmelo92118, line 4954: Unavailable.

IceCodeNew:
忙，晚上看

data/category-porn
IceCodeNew: > ```
> bfulibl.net
> fulibl.cc
> fulibl.com
> fulibl.net
> fulibl.org
> fulibl.xyz
> ifulibl.net
> ```
> 
> Sometimes, it will add an alphabet before `fulibl`. Actually, I wanna use the syntax `keyword:`, but I am afraid it will match the domain which should not be matched. And no one uses it in this repository, so I choose the `regexp:` to realize it.

---

In that case, I will suggest the following regexp:
```
regexp:^([a-zA-Z0-9]*)fulibl\.(cc|com|net|org|xyz)$
```

it makes sense to cover some domains that are not being used for now, but the previous one had gone too far...

IceCodeNew (approved):

PR #1508, opened 2023-02-27
Create ir-ads

mmrabbani:
I have created list of ad domains in Iran in https://github.com/v2fly/domain-list-community/pull/1376
That pull request had license conflict and closed unmerged https://github.com/v2fly/domain-list-community/pull/1376#issuecomment-1436638351
So I created this list by referencing another list with MIT license. So there would not be license conflict anymore.


rootmelo92118 (approved):

IceCodeNew:
Thanks again for showing interest in contributing to this project. I appreciate it.
But here I can not help to raise the same question: Why would we want to have a copy from other projects?
You see, the license issue is not the main idea I declined the previous PR #1376. It is just not wise to maintain two copies in two separate projects.
It is not practical to consistently update the local copy, combine with the exact labor and risk for solving syntax differences between two projects, I can not convince myself to accept such a PR.

Is there anything preventing you from trying the workaround I mentioned in the previous PR? IMO you really should consider giving the following project a try.
Any content you would like to merge into this project can be implemented here, using the Linux shell commands.
https://github.com/Loyalsoldier/domain-list-custom/blob/cc8590a9489c8f5fa484f80cd32beb1449cb10e8/.github/workflows/build.yml#L42

IceCodeNew (changes requested):
Thanks again for showing interest in contributing to this project. I appreciate it.
But here I can not help to raise the same question: Why would we want to have a copy from other projects?
You see, the license issue is not the main idea I declined the previous PR #1376. It is just not wise to maintain two copies in two separate projects.
It is not practical to consistently update the local copy, combine with the exact labor and risk for solving syntax differences between two projects, I can not convince myself to accept such a PR.

Is there anything preventing you from trying the workaround I mentioned in the previous PR? IMO you really should consider giving the following project a try.
Any content you would like to merge into this project can be implemented here, using the Linux shell commands.
https://github.com/Loyalsoldier/domain-list-custom/blob/cc8590a9489c8f5fa484f80cd32beb1449cb10e8/.github/workflows/build.yml#L42

rootmelo92118:
In my opinion, this PR can help us make this repository more complete. But, the problem we will face is this category only can be updated by ourselves. The maintainer and the contributor cannot check the remote project whether is updated or not every day. However, I can understand a lot of users don't have enough ability to generate the custom `geosite.dat` and use it. So, I will suggest you maybe you can try to include that in [here](https://github.com/Loyalsoldier/v2ray-rules-dat). This project will generate the custom `geosite.dat` every morning. When it generates the new file, the domain's data will be updated automatically.

mmrabbani:
@IceCodeNew 

> Why would we want to have a copy from other projects?

I know contributing without using other projects has some benefits, but I (and no one else) cannot find more than 300 domains myself. this list is created by years of contributing.

> It is just not wise to maintain two copies in two separate projects.

> The maintainer and the contributor cannot check the remote project whether is updated or not every day.

I can write at the first line of file that this list is not maintained from original project continuously. It is copied one time in 02/2023 and will be maintained by "domain-list-community" maintainers themselves after that. This list is created in years and does not updated frequently.

> Is there anything preventing you from trying the workaround I mentioned in the previous PR? IMO you really should consider giving the following project a try.

The reason I insist on contributing in this project and not Loyalsoldier project is that this project is used by many famous v2ray clients and server apps and users will not need to import a file to use this list. So users can use this list easily. Users that block list of ad domains before (block:category-ads-all) don't even need to do anything and benefit from this list automatically.

IceCodeNew:
> but I (and no one else) cannot find more than 300 domains myself. this list is created by years of contributing.

That is not true. Most of the content I introduced to this project is based on what I accessed. As for now, I may have contributed about 300 domains ;-)
Submitting a PR is meant to, and should only be meant to solve a specific problem of the use case that contributors concern about. Trying to make a category _perfect_ by taking other projects' work does not make sense to me, and it is overkill for solving your own problem.
// Not to mention that for bypassing censorship, the quality of the ads list does not matter at all. So please try not to be fixed on making a category complete and perfect.

> This list is created in years and does not updated frequently.

It does not matter. As I said before.

> I can write at the first line of file that this list is not maintained from original project continuously. It is copied one time in 02/2023 and will be maintained by "domain-list-community" maintainers themselves after that.

It does not solve the problem.

> The reason I insist on contributing in this project and not Loyalsoldier

No, I did not recommend you to contribute to the https://github.com/Loyalsoldier/domain-list-custom project, the owner of that project would probably close your PR immediately. _You need to fork that repo._

> The reason I insist on contributing in this project and not Loyalsoldier project is that this project is used by many famous v2ray clients and server apps and users will not need to import a file to use this list. So users can use this list easily. Users that block list of ad domains before (block:category-ads-all) don't even need to do anything and benefit from this list automatically.

Sure it is. Though you may notice that people who lived outside of Iran could have completely different needs, and everyone wants the best out-of-the-box experience for them. How can a project handle various needs, especially when some of them may conflict with others?
I will approve PRs as long as they did not trigger me to raise the benefit-cost ratio question. And I am afraid the decision is final, this PR will not get merged.
You have to make changes in a fork if you really need those ads-blocking lists.

PR #1554, opened 2023-03-24
Update apple

UptonEdward:
Add a CDN domain name that being used for the `netQuality` speedtest command.


rootmelo92118:
Yes, you should. The domain `aaplimg.com` currently belongs to apple.
截圖 2023-03-25 上午8 35 49

PR #1555, opened 2023-03-24
Update private

francis-zhao:
Added
- Added `full:asusrouter.com`/`full:www.asusrouter.com` used by new firmware of ASUS routers.
- Added `full:tplogin.cn` used by TP-Link routers of mainland China version.
- Added `plex.direct` used by Plex Media Server on NAS/HTPC.

Changed
- Reclassified `miwifi.com` to Xiaomi routers and changed to `full:miwifi.com`/`full:www.miwifi.com`.
- Reclassified `routerlogin.com` to Netgear routers and changed to `full:routerlogin.com`/`full:www.routerlogin.com`.
- Modified some comments.

### DIFF
--- a/data/private
+++ b/data/private
@@ -126,14 +126,16 @@ full:instant.arubanetworks.com
 full:setmeup.arubanetworks.com
 
 # ASUS router
+full:asusrouter.com
 full:router.asus.com
-asusrouter.com
+full:www.asusrouter.com
 
 # KDE
 networkcheck.kde.org
 
 # Netgear router
-routerlogin.com
+full:routerlogin.com
+full:www.routerlogin.com
 
 # Other router
 hiwifi.com
@@ -159,7 +161,8 @@ msftconnecttest.com
 msftncsi.com
 
 # Xiaomi router
-miwifi.com
+full:miwifi.com
+full:www.mifiwi.com
 
 # ZTE router
 zte.home
--- a/data/private
+++ b/data/private
@@ -154,7 +154,7 @@ tendawifi.com
 
 # TP-Link router
 tplinkwifi.net
-tplogin.cn
+full:tplogin.cn
 
 # Windows
 msftconnecttest.com
--- a/data/private
+++ b/data/private
@@ -126,6 +126,7 @@ full:instant.arubanetworks.com
 full:setmeup.arubanetworks.com
 
 # ASUS router
+# Reference: https://www.asus.com/support/FAQ/1005263/
 full:asusrouter.com
 full:router.asus.com
 full:www.asusrouter.com
@@ -134,6 +135,8 @@ full:www.asusrouter.com
 networkcheck.kde.org
 
 # Netgear router
+# Reference: https://kb.netgear.com/27199/I-can-t-access-my-router-what-do-I-do
+# https://www.netgear.com/business/services/aplogincom/
 full:routerlogin.com
 full:www.routerlogin.com
 
@@ -145,6 +148,10 @@ peiluyou.com
 phicomm.me
 router.ctc
 
+# Plex Media Server
+# Reference: https://support.plex.tv/articles/206225077-how-to-use-secure-server-connections/
+plex.direct
+
 # QQ SSO for quick login
 localhost.sec.qq.com
 localhost.ptlogin2.qq.com
@@ -153,6 +160,7 @@ localhost.ptlogin2.qq.com
 tendawifi.com
 
 # TP-Link router
+# Reference: https://resource.tp-link.com.cn/m/productClass/product_document?id=1655112502304621
 tplinkwifi.net
 full:tplogin.cn
 
@@ -161,6 +169,7 @@ msftconnecttest.com
 msftncsi.com
 
 # Xiaomi router
+# Reference: https://www1.miwifi.com/miwifi_faq.html
 full:miwifi.com
 full:www.mifiwi.com
 
--- a/data/private
+++ b/data/private
@@ -126,7 +126,7 @@ full:instant.arubanetworks.com
 full:setmeup.arubanetworks.com
 
 # ASUS router
-router.asus.com
+full:router.asus.com
 asusrouter.com
 
 # KDE
--- a/data/private
+++ b/data/private
@@ -126,22 +126,24 @@ full:instant.arubanetworks.com
 full:setmeup.arubanetworks.com
 
 # ASUS router
-full:router.asus.com
+router.asus.com
+asusrouter.com
 
 # KDE
 networkcheck.kde.org
 
+# Netgear router
+routerlogin.com
+
 # Other router
 hiwifi.com
 leike.cc
-miwifi.com
 my.router
 peiluyou.com
 phicomm.me
 router.ctc
-routerlogin.com
 
-# QQ Loopback
+# QQ SSO for quick login
 localhost.sec.qq.com
 localhost.ptlogin2.qq.com
 
@@ -150,10 +152,14 @@ tendawifi.com
 
 # TP-Link router
 tplinkwifi.net
+tplogin.cn
 
 # Windows
 msftconnecttest.com
 msftncsi.com
 
-# 中興路由器
+# Xiaomi router
+miwifi.com
+
+# ZTE router
 zte.home


data/private
```diff
@@ -126,22 +126,24 @@ full:instant.arubanetworks.com
 full:setmeup.arubanetworks.com
 
 # ASUS router
-full:router.asus.com
+router.asus.com
```
IceCodeNew, line 129: Did not get it. Why?

data/private
```diff
@@ -126,22 +126,24 @@ full:instant.arubanetworks.com
...
+asusrouter.com
 
 # KDE
 networkcheck.kde.org
 
+# Netgear router
+routerlogin.com
```
IceCodeNew, line 136: Use the `full:` prefix whenever you can to avoid mismatch.

data/private
```diff
@@ -126,22 +126,24 @@ full:instant.arubanetworks.com
 full:setmeup.arubanetworks.com
 
 # ASUS router
-full:router.asus.com
+router.asus.com
+asusrouter.com
```
IceCodeNew, line 130: Use the `full:` prefix whenever you can to avoid mismatch.

data/private
```diff
@@ -150,10 +152,14 @@ tendawifi.com
 
 # TP-Link router
 tplinkwifi.net
+tplogin.cn
```
IceCodeNew, line 155: Use the `full:` prefix whenever you can to avoid mismatch.

data/private
```diff
@@ -150,10 +152,14 @@ tendawifi.com
...
 
 # Windows
 msftconnecttest.com
 msftncsi.com
 
-# 中興路由器
+# Xiaomi router
+miwifi.com
```
IceCodeNew, line 162: Use the `full:` prefix whenever you can to avoid mismatch.

IceCodeNew (changes requested):

data/private
francis-zhao: Changed back to `full`

data/private
francis-zhao: Using `full:` can't match `www.routerlogin.com`, so just use `domain:routerlogin.com` for all situations.

data/private
francis-zhao: Using `full:` can't match `www.asusrouter.com`, so just use `domain: asusrouter.com` for all situations.

data/private
francis-zhao: Same as above.

data/private
francis-zhao: Same as above.

data/private
IceCodeNew: > It doesn't matter if the url has a `www` prefix or not, both can be accessed. Using `full:` can't match `www.asusrouter.com`, so just use `domain:asusrouter.com` for all situations, and this domain only used for router management.

Yes, it does matter. This is the `private` category and currently we are abusing it.
So I want only the necessary domains to be included in this category, not those we can not determine whether it is used for intranet access or not.

IceCodeNew: > It doesn't matter if the url has a `www` prefix or not, both can be accessed. Using `full:` can't match `www.asusrouter.com`, so just use `domain:asusrouter.com` for all situations, and this domain only used for router management.

Yes, it does matter. This is the `private` category and currently we are abusing it.
So I want only the necessary domains to be included in this category, not those we can not determine are used for intranet access.

IceCodeNew: > It doesn't matter if the url has a `www` prefix or not, both can be accessed. Using `full:` can't match `www.asusrouter.com`, so just use `domain:asusrouter.com` for all situations, and this domain only used for router management.

Yes, it does matter. This is the `private` category and currently we are abusing it.
So I want only the necessary domains to be included in this category, not those we can not determine are used for intranet access.

data/private
IceCodeNew: > The url on the label of router is `routerlogin.com`, using `full:` can't match `www.routerlogin.com`, which is accessible too, so just use `domain:routerlogin.com` for all situations since we don't know which url user may input.

Then why not just let it be `full:www.routerlogin.com`

francis-zhao: Added `full:routerlogin.com` and `full:www.routerlogin.com`.

data/private
francis-zhao: Added `full:asusrouter.com` and `full:www.asusrouter.com`.

data/private
francis-zhao: This domain has no A/AAAA/CNAME records via `nslookup` and can't get echo from `ping`, only used for intranet, so I leave it untouched.

IceCodeNew: > This domain has no A/AAAA/CNAME records via `nslookup` and can't get echo from `ping`, only used for intranet, so I leave it untouched.

We should avoid mismatches whenever we can. What if someone, who owns this domain, decides to make use of a sub-domain under it in the future?
Making changes based on solid proofs would be better.

data/private
```diff
@@ -126,22 +126,26 @@ full:instant.arubanetworks.com
...
+full:asusrouter.com
 full:router.asus.com
+full:www.asusrouter.com
 
 # KDE
 networkcheck.kde.org
 
+# Netgear router
+full:routerlogin.com
+full:www.routerlogin.com
```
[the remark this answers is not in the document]
IceCodeNew: > If your router isn't NETGEAR, then just visit this domain, it will redirect to the setup page that shows how to setup your router correctly since this domain didn't filtered by the router.
> 
> If your domain is NETGEAR, visit this domain will show the management page.

You can just show me the evidence, you know.
Since this is reviewing and I do not have the same knowledge on the background of this PR as you have. It is up to you to knowledge me till I can see the benefits of doing so, or I will not pull the trigger.
No source, no go.

data/private
francis-zhao: https://resource.tp-link.com.cn/m/productClass/product_document?id=1655112502304621

Added `full:tplogin.cn`.

data/private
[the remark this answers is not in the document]
francis-zhao: I got you, I will append references next time.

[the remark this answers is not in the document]
francis-zhao: https://kb.netgear.com/27199/I-can-t-access-my-router-what-do-I-do

This page gives you proof, and my last router is NETGEAR, the url on the label is `routerlogin.com`.

This page also shows another domain `routerlogin.net`, but not used by my model. My old NETGEAR router has been sold, and now I can’t verify whether this is also available, so I didn’t add it.

francis-zhao:
I just added another domain `plex.direct` used by Plex Media Server.

Reference: https://support.plex.tv/articles/206225077-how-to-use-secure-server-connections/
Just navigate to the last part of this article or search the keyword `plex.direct`.

data/private
```diff
@@ -126,34 +126,52 @@ full:instant.arubanetworks.com
...
+full:tplogin.cn
 
 # Windows
 msftconnecttest.com
 msftncsi.com
 
-# 中興路由器
+# Xiaomi router
+# Reference: https://www1.miwifi.com/miwifi_faq.html
+full:miwifi.com
```
rootmelo92118, line 173: According to your reference data, I only see the domain name `www.miwifi.com` is used for the manage panel. Do you have any evidence to prove this domain is used in `private`?

francis-zhao: The manage panel is the router's inner page, why the domain used for it is not `private`? All other domains used for routers in this `private` list are just substitutions for local ip such as `192.168.1.1`.

By default, I input `miwifi.com` or `www.miwifi.com`, it opens Xiaomi router's admin page as same as I input `192.168.31.1`, this page is hosted on the router of my local network.

image

image

IceCodeNew:
I had roughly reviewed this PR and came up with several questions that I want to confirm.
I may draw the conclusion in late today or not, sorry for the inconvenience.

data/private
rootmelo92118: 截圖 2023-03-28 上午10 17 59

Ok, I have no problem.

IceCodeNew:
The newest ETA is the coming Saturday night, thanks for everyone being patient.

data/private
```diff
@@ -126,34 +126,52 @@ full:instant.arubanetworks.com
 full:setmeup.arubanetworks.com
 
 # ASUS router
+# Reference: https://www.asus.com/support/FAQ/1005263/
+full:asusrouter.com
```
IceCodeNew, line 130: Can not see the reason why we should introduce this domain into this category.

data/private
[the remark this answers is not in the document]
IceCodeNew: I think it is OK to not include the domain `routerlogin.net` for now. I insist only domains that could be verified be included in this category.

data/private
```diff
@@ -126,34 +126,52 @@ full:instant.arubanetworks.com
...
 my.router
 peiluyou.com
 phicomm.me
 router.ctc
-routerlogin.com
 
-# QQ Loopback
+# Plex Media Server
+# Reference: https://support.plex.tv/articles/206225077-how-to-use-secure-server-connections/
+plex.direct
```
IceCodeNew, line 153: why not `full:plex.direct`?
// the `.direct` does not look like a valid TLD for the internet, but only for now.

IceCodeNew (changes requested):

data/private
francis-zhao: image

As shown in screenshot above, Plex uses individual subdomains under the `plex.direct` for each server instance, not this domain directly.

data/private
francis-zhao: https://user-images.githubusercontent.com/37755859/229323400-a2d20cd9-a686-4052-955a-26b3da6724a1.mp4

When ASUS started to use new domain `www.asusrouter.com` for admin page of the new firmware version of routers, the new without the `www` prefix and the legacy `router.asus.com` are also supported.

Now these two are aliases of the new domain, and inner service of the router will automatically redirect (not rewrite) to the new domain. 

image

`www.asusrouter.com` is resolved to `192.168.50.1`.
`asusrouter.com` and `router.asus.com` are rebinded to `10.0.0.1` and handled by inner service for further process.

IceCodeNew (approved):

IceCodeNew:
Thanks for your contribution!

Issue #1561, opened 2023-03-26
Solve multiple root node problem and decrease it


rootmelo92118:
I write a script to find the categories which are not included by other categories. And I get this result.
```
category-finance
adobe-activation
category-network-security
speedtest
pandanet
test
category-pt
category-ads-all
cn
category-public-tracker
cowlevel
biliplus
tld-opennic
chemequations
dslreports
private
tor
apple-update
category-enhance-gaming
showtimeanytime
geolocation-!cn
category-ir
downloadly
category-ru-gov
hubblephone
```
Some categories should not be included by others. That is according to the discussion with other collaborators so I won't touch them.
Besides, I will create some pull requests to include them in others.

IceCodeNew:
Would you mind me postponing the review till April? It is indeed a good idea, but I think I can only make a call when I have time to review all of this; Digesting them one by one may lead to misunderstanding and a lapse in judgment.

BTW, you did it very well in splitting the changes into several small PRs.
However, you could make it better by just opening this issue, and waiting till you have had a discussion with people who is likely to review the PRs.
You can start your work after you have put some effort to make sure that any overlooking has been caught in advance, and reviewers show interest in your proposal.
@rootmelo92118

rootmelo92118:
Oh, I am sorry to make you misunderstand. For these categories, my strategy is to include them in existing categories. If you think it should not belong to some categories, welcome to make a discussion.
@IceCodeNew

IceCodeNew:
The newest ETA is the coming Saturday night, thanks for everyone being patient.

IceCodeNew:
Finally, I can have some rest...

PR #1563, opened 2023-03-26
Solve multiple root node problem and decrease it (geolocation-cn)

rootmelo92118:
#1561


data/geolocation-cn
```diff
@@ -147,6 +147,9 @@ include:category-media-cn
 # Organizations
 include:category-hospital-cn
 
+# Network Security
+include:category-network-security
```
IceCodeNew, line 151: It is not guaranteed that only Chinese sites will be included in this category.
A sloppy solution is to rename it to `category-network-security`, hoping the renaming would not cause any other problems.

IceCodeNew (changes requested):

data/geolocation-cn
rootmelo92118: What is the name you want to change?

data/category-network-security-cn
IceCodeNew, line 1: 我觉得有必要把 `@cn` 的属性标记删掉。

IceCodeNew (changes requested):

IceCodeNew (approved):

PR #1566, opened 2023-03-26
Solve multiple root node problem and decrease it (geolocation-!cn)

rootmelo92118:
#1561

### DIFF
--- a/data/geolocation-!cn
+++ b/data/geolocation-!cn
@@ -170,6 +170,7 @@ include:adguard
 include:aptoide
 include:arphic
 include:authy
+include:biliplus
 include:bitly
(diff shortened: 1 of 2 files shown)


data/geolocation-!cn
```diff
@@ -252,6 +252,8 @@ include:zeplin
 include:zoho
 include:zoom
 
+biliplus.com
```
IceCodeNew, line 255: 加一行分类注释吧

IceCodeNew (changes requested):

data/geolocation-!cn
rootmelo92118: done

PR #1567, opened 2023-03-26
Solve multiple root node problem and decrease it (geolocation-!cn)

rootmelo92118:
#1561

### DIFF
--- a/data/geolocation-!cn
+++ b/data/geolocation-!cn
@@ -86,7 +86,6 @@ include:betterexplained
 include:brilliant
 include:category-scholar-!cn
 include:cern
-include:chemequations
 include:chinapower
 include:cnet
 include:code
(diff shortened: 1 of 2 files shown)


data/geolocation-!cn
```diff
@@ -115,6 +115,8 @@ include:udacity
 include:udemy
 include:w3schools
 
+chemequations.com
```
IceCodeNew, line 118: 给这个域名加一行注释，写上分类吧。

IceCodeNew (changes requested):

data/geolocation-!cn
rootmelo92118: done

IceCodeNew (approved):

PR #1570, opened 2023-03-26
Solve multiple root node problem and decrease it (category-ir-tech)

rootmelo92118:
#1561

### DIFF
--- a/data/category-ir-tech
+++ b/data/category-ir-tech
@@ -1,5 +1,7 @@
 # Iranian technology websites
 
+include:downloadly
+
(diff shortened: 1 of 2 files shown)


nickaein:
It doesn't seem to be listed in other categories, so we probably should add it to `category-ir-tech`.

nickaein:
To be clear, I'm up for deleting `downloadly` file and adding an `downloadly.ir` entry to `cateogry-ir-tech` instead.

IceCodeNew (approved):

PR #1596, opened 2023-04-03
Add neshanmap.ir, tapsell.ir, yektanet.com

keikhosrow:


data/category-ir-tech
```diff
@@ -26,4 +27,6 @@ shatel.ir
 soft98.ir
 sotoon.ir
 taliya.ir
+tapsell.ir
```
IceCodeNew, line 30: if this domain belongs to an e-commerce site, we should move it to a category named `category-ir-e-commerce` or something.

IceCodeNew (changes requested):

data/category-ir-tech
```diff
@@ -26,4 +27,6 @@ shatel.ir
...
 tci.ir
+yektanet.com
```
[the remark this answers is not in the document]
rootmelo92118: 截圖 2023-04-16 下午4 11 00
This website is about advertising.

data/category-ir-tech
rootmelo92118: 截圖 2023-04-16 下午4 12 38
This website is also about advertising.

IceCodeNew:
There are 3 actions we can take, which one would you prefer?
1. do not include `tapsell.ir` & `yektanet.com`
2. move these two domains into a category like `category-ir-ads`, which I consider the worst move. Blocking ads/trackers based on domains is not only inaccurate, but also a nightmare for reviewing and maintaining, and a fuse for the copyleft problems
3. keep these domains as they could be considered as _tech companies_

rootmelo92118:
> There are 3 actions we can take, which one would you prefer?
> 
> 1. do not include `tapsell.ir` & `yektanet.com`
> 2. move these two domains into a category like `category-ir-ads`, which I consider the worst move. Blocking ads/trackers based on domains is not only inaccurate, but also a nightmare for reviewing and maintaining, and a fuse for the copyleft problems
> 3. keep these domains as they could be considered as _tech companies_
> 
> @rootmelo92118

@IceCodeNew I am prefer the third action. Do you have any opinion?

keikhosrow:
Hello, due to the fact that the biggest advertising platform in Iran are four websites: Yektanet, Tapsel, Saba Vision and Tavus, I think the third option is better for now.

rootmelo92118 (approved):

IceCodeNew:
> Hello, due to the fact that the biggest advertising platform in Iran are four websites: Yektanet, Tapsel, Saba Vision and Tavus, I think the third option is better for now.

I am OK with that. Thanks for feeding back.

IceCodeNew (approved):

PR #1598, opened 2023-04-04
Add YouTube ads

AC-Lover:


EpLiar:
Thank you for contributing! But it's too large to review, can you provide the source of these domains?

rootmelo92118:
> Thank you for contributing! But it's too large to review, can you provide the source of these domains?

I think we can reference the conclusion of #1058.

PR #1602, opened 2023-04-07
Update category-porn

bijoni:

PR #1603, opened 2023-04-07
Update category-porn

IceCodeNew:
fix #1602 
@bijoni @rootmelo92118 PTAL

### DIFF
--- a/data/category-porn
+++ b/data/category-porn
@@ -3381,11 +3381,6 @@ nudeteen.org
 nudeteenboys.net
 nudeteenladies.com
 nudevietnam.com
-nudevista.best
-nudevista.club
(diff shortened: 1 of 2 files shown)


rootmelo92118:
Although they are using the same domain name, they are not the same website. They should not include in `nudevista`.

rootmelo92118 (changes requested):
Although they are using the same domain name, they are not the same website. They should not include in `nudevista`.

IceCodeNew:
> Although they are using the same domain name, they are not the same website. They should not include in `nudevista`.

I saw the different designs, and still jumped to the conclusion since they were using the same slogan. It may not be the case here.

PR #1623, opened 2023-04-21
add divar cdn

nickaein:


rootmelo92118:
Do you have any source?

nickaein:
Hi Kitayama, Sure! 

By browsing the [Divar website](https://divar.ir), you can verify that the static resources are being served at `divarcdn.com` domain.

Also, [DNS records of divarcdn.com](https://www.nslookup.io/domains/divarcdn.com/dns-records/) points to IP addresses that their geoloction are inside Iran.

P.S. Divar is the [most popular ads platform in Iran with ~53 million users](https://en.wikipedia.org/wiki/Divar_(website)), so it would be helpful to cover its all domains.

rootmelo92118 (approved):

rootmelo92118:
Thanks for your contribution.

PR #1658, opened 2023-05-12
Added huangyuhui.net to geolocation-cn

Nep-Timeline:


IceCodeNew:
I think it would be better to put that domain under a category named `category-blog-cn`, then you can include it in this `geolocation-cn` category.

IceCodeNew (changes requested):
I think it would be better to put that domain under a category named `category-blog-cn`, then you can include it in this `geolocation-cn` category.

Nep-Timeline:
> I think it would be better to put that domain under a category named `category-blog-cn`, then you can include it in this `geolocation-cn` category.

#1659 like this?

IceCodeNew:
suppressed by #1659

Issue #1337, opened 2022-12-01
Request iranian domain list


debMan:
Can anybody help and make a list obtained from the official regulations of Iran?

https://eservices.ito.gov.ir/Page/IPList

rootmelo92118:
Could you please upload this list here? It looks like reject the connection outside Iran.

IceCodeNew:
> Can anybody help and make a list obtained from the official regulations of Iran?
> 
> [eservices.ito.gov.ir/Page/IPList](https://eservices.ito.gov.ir/Page/IPList)

Would you please have a look at #1294?
If there are domains that were missing in this PR, you can submit another PR for amendment.

debMan:
Dear @IceCodeNew @rootmelo92118, you both are correct. The domain is accessible only inside Iran.
It contains a table with paginations, so I have to crawl them and add them as text.
![image](https://user-images.githubusercontent.com/12995147/205301723-f0a1bacc-1bb0-4918-9a79-afcbb37d066c.png)

> Would you please have a look at #1294? If there are domains that were missing in this PR, you can submit another PR for amendment.
> 
The PR is correct to some extent, but there are more domains which are not been added yet.

IceCodeNew:
> > Would you please have a look at #1294? If there are domains that were missing in this PR, you can submit another PR for amendment.
> 
> The PR is correct to some extent, but there are more domains which are not been added yet.

It would be nice if you could submit a PR for the missing domains since I and @rootmelo92118 can only make the PR based on a screenshot, which is pretty inconvenient.

rootmelo92118:
@debMan Is it?
[Export-14010911183802.xls](https://github.com/v2fly/domain-list-community/files/10142439/Export-14010911183802.xls)

debMan:
> > > Would you please have a look at #1294? If there are domains that were missing in this PR, you can submit another PR for amendment.
> > 
> > 
> > The PR is correct to some extent, but there are more domains which are not been added yet.
> 
> It would be nice if you could submit a PR for the missing domains since I and @rootmelo92118 can only make the PR based on a screenshot, which is pretty inconvenient.

Sure, thanks. I'll work on it ASAP.

debMan:
> @debMan Is it? [Export-14010911183802.xls](https://github.com/v2fly/domain-list-community/files/10142439/Export-14010911183802.xls)

Does `geopip` of Iran covers all iranian non `.ir` domains?

rootmelo92118:
[...]

PR #1661, opened 2023-05-13
All domains that have access point in Iran

EryX0:
addresses #1337
it differs from category-ir which only contains domains that won't open with foreign ip's, this one is all of the websites that are hosted/have access point in Iran.


rootmelo92118:
https://github.com/rootmelo92118/v2ray-rules-dat

PR #1673, opened 2023-05-20
add `mdn.dev`

yin1999:
MDN is now using `mdn.dev` to serve some archived media files, add this to the "mdn" entry.

- mdn/translated-content#13291
- mdn/yari#8873

### DIFF
--- a/data/mdn
+++ b/data/mdn
@@ -1,5 +1,6 @@
 developer.allizom.org
 developer.mozilla.org
+mdn.dev
 mdn.mozit.cloud
 
 full:interactive-examples.mdn.mozilla.net


EpLiar (approved):

data/mdn
```diff
@@ -1,5 +1,6 @@
 developer.allizom.org
 developer.mozilla.org
+mdn.dev
 mdn.mozit.cloud
```
caugner, line 4: @yin1999 FYI

- We will stop using `mdn.mozit.cloud` soon.
- We stopped using `mdn.mozillademos.org` (it redirects to `mdn.dev`).
- You could add `mdn.mozilla.net` and `mdn.allizom.net`, because there's also `live-samples.mdn.mozilla/allizom.net` (`allizom.net` is rather new).

PR #1686, opened 2023-05-28
Create trackers

mmrabbani:
This list contains top web trackers. blocking this list can help privacy oriented users.

### DIFF
--- a/data/trackers
+++ b/data/trackers
@@ -1,0 +1,5 @@
+appcenter.ms
+app-measurement.com
+firebase.io
+crashlytics.com
+google-analytics.com
--- a/data/trackers
+++ b/data/trackers
@@ -1,5 +1,5 @@
-appcenter.ms
 app-measurement.com
-firebase.io
+appcenter.ms
 crashlytics.com
+firebase.io
 google-analytics.com


mmrabbani:
this list can be add to "category-ads-all". But I didn't add this. because I think maintainers may not like this

data/trackers
```diff
@@ -0,0 +1,5 @@
+app-measurement.com
+appcenter.ms
```
IceCodeNew, line 2: this will match domains that are not for tracking users.

data/trackers
```diff
@@ -0,0 +1,5 @@
+app-measurement.com
```
IceCodeNew, line 1: already exist.

data/trackers
```diff
@@ -0,0 +1,5 @@
...
+appcenter.ms
+crashlytics.com
```
IceCodeNew, line 3: already exist.

data/trackers
```diff
@@ -0,0 +1,5 @@
...
+firebase.io
+google-analytics.com
```
IceCodeNew, line 5: already exist.

IceCodeNew (changes requested):

PR #1725, opened 2023-06-22
Add zxinc.org

herberteuler:
https://ip.zxinc.org/ipquery/

This site displays IPv4 and IPv6 addresses of the viewer simultaneously, and one of the two addresses can be the client's local address, causing a physical location clash.

### DIFF
--- a/data/zxinc
+++ b/data/zxinc
@@ -1,0 +1,1 @@
+zxinc.org


rootmelo92118:
Do you think which category it should belong to?

herberteuler:
Okay, I was not aware of the categories.
I will refile another PR.

PR #1737, opened 2023-07-05
Update bytedance

coderbean:
add 3 new inner domain

### DIFF
--- a/data/bytedance
+++ b/data/bytedance
@@ -33,8 +33,6 @@ ibytedtos.com @!cn
 ibyteimg.com @!cn
 pstatp.com
 snssdk.com
-tiktokd.org
-tiktokd.net
 wtturl.cn
 zijieapi.com
 zijiecdn.com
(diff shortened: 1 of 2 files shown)


data/bytedance
```diff
@@ -32,6 +33,8 @@ ibytedtos.com @!cn
 ibyteimg.com @!cn
 pstatp.com
 snssdk.com
+tiktokd.org
+tiktokd.net
```
rootmelo92118, line 37: Line 36 and 37. If they are only used on tiktok, it should be included in `tiktok`

rootmelo92118 (changes requested):

data/bytedance
rootmelo92118: @coderbean Hello?

coderbean: > @coderbean Hello?
Hi


> Line 36 and 37. If they are only used on tiktok, they should be moved to `tiktok`

ok, I'll do it.

rootmelo92118 (approved):

PR #1760, opened 2023-07-24
Update category-porn

Toad-stool:


data/category-porn
```diff
@@ -43,6 +43,7 @@ include:netflav
 include:nudevista
 include:picacg
 include:pornpros
+include:pornhub
```
rootmelo92118, line 46: Pornhub had already been included in mindgeek-porn and the mindgeek-porn also had been included in category-porn.

rootmelo92118: Please remove it

rootmelo92118: @Toad-stool

PR #1789, opened 2023-07-31
Rename twitter to X

OPPO9008:
twitter 改名叫X 了


Loyalsoldier:
I think it's better to add a new category called `x` and include `twitter` in it. V2Ray fails to start if no `geosite:twitter` exists.

PR #1805, opened 2023-08-17
Remove CN battle link

fwqaaq:
These cn website all redirect to `www.163.com`, and no necessary to reserve them.


rootmelo92118:
Could you explain what is your purpose?

fwqaaq:
> Could you explain what is your purpose?

These urls have been no longer valid when I checked out them. Maybe battle wouldn't use these urls in the feature when it come back cn?

rootmelo92118 (approved):

PR #1847, opened 2023-10-01
Update `tld-!cn`, Create `tld-ir`, Update `category-ir`

demarcush:
Iraq typo


rootmelo92118:
If the tld will not increase in future. I think we don't need to create category `tld-ir`.

rootmelo92118 (approved):

PR #1848, opened 2023-10-01
Update `category-anticensorship` and `category-browser-!cn`

demarcush:


data/category-anticensorship
```diff
@@ -1,8 +1,13 @@
+include:category-ipfs
```
rootmelo92118, line 1: Why?

rootmelo92118 (approved):

PR #1860, opened 2023-10-13
update url for suishouji and others

rp-hello:


rootmelo92118:
Please separate this PR into two parts. I don't have any opinion of xiaohongshu. But another part I think we need to do something.

rootmelo92118 (changes requested):
Please separate this PR into two parts. I don't have any opinion of xiaohongshu. But another part I think we need to do something.

PR #1862, opened 2023-10-15
Update google-ads

xkww3n:
Google Analytics, Tag Manager and Firebase aren't AD services.


Loyalsoldier:
I think it would be better to keep them.

xkww3n:
> I think it would be better to keep them.

There do have some people need these tracking services' domains, but they shouldn't be categorized as AD services. Maybe it's better to have another list like google-tracking instead of putting these into AD list.

PR #1889, opened 2023-10-29
Update apple

peeweep:


data/apple
```diff
@@ -800,6 +808,7 @@ full:app-site-association.cdn-apple.com @cn
 full:appldnld.apple.com @cn
 full:appldnld.g.aaplimg.com @cn
 full:appleid.cdn-apple.com @cn
+full:apps.apple.com @cn
```
rootmelo92118, line 811: Could you provide any evidence of it?

database64128: Unless #503 is no longer the case, this particular domain should not be included.

Maybe keep it as a comment to remind future maintainers?

rootmelo92118: I do agree.

rootmelo92118 (approved):

PR #1899, opened 2023-11-05
Update openai

qifeng2233:
add chatgpt.com


data/openai
rootmelo92118, line 1: sort it.

rootmelo92118 (changes requested):

qifeng2233 (commented):
change sort

PR #1903, opened 2023-11-07
Update cnbeta

percyhall:
Add cnbeta.com.tw


Loyalsoldier:
Please create a new file called `cnbeta` and move those related domains into it, then include it into the category.

PR #1909, opened 2023-11-11
Create cnbeta and include into category-media

leewis101:


Loyalsoldier:
Please include into `category-media-cn`, not `category-media`

PR #1919, opened 2023-11-18
Update category-porn

chunyunghsu:
The reason for removing the domain is they are no longer available.


rootmelo92118:
Thanks you for get rid of it. Would you mind help me to do something? Could you help us to remove all the domain which is unavailable or the porn content is no longer available in the website?

chunyunghsu:
Well. Maybe I can do this if I have time.

rootmelo92118 (approved):

PR #2000, opened 2023-12-31
Add httpdns

chika0801:
from

github.com/VirgilClyne/GetSomeFries/blob/main/ruleset/HTTPDNS.Block.yaml

github.com/MetaCubeX/meta-rules-dat/blob/sing/geo/geosite/httpdns.json


rootmelo92118 (changes requested):
This category should be call `category-httpdns` and please include this category to relevant category.

PR #2139, opened 2024-04-28
Update CN related domains

xkww3n:
- Add `!cn` tags for overseas domains in serval cn banks.
- Remove terminated TLDs in tld-!cn.


data/tld-!cn
rootmelo92118, line 1: lines 41, 77, 151 are not be terminated, but it never has the Registry Provider..

And I hope you can provide the evdience of the domains that you removed?

rootmelo92118 (changes requested):

rootmelo92118 (approved):

PR #2147, opened 2024-05-04
Update category-ads

rootmelo92118:


Loyalsoldier:
According  to https://github.com/v2fly/domain-list-community/pull/2123 , `facebook-ads` and `meta-ads` should not be included.

PR #2153, opened 2024-05-06
add some domains of apis used by Clubhouse

hanyujie2002:
Without these apis, clubhouse can't work well

### DIFF
--- a/data/clubhouse
+++ b/data/clubhouse
@@ -4,3 +4,4 @@ joinclubhouse.com
 agora.io
 branch.io
 sentry.io
+amazonaws.com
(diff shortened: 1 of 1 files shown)


data/clubhouse
```diff
@@ -1,3 +1,6 @@
 clubhouse.com
 clubhouseapi.com
 joinclubhouse.com
+agora.io
+branch.io
+sentry.io
```
hanyujie2002, line 6:
```suggestion
sentry.io
amazonaws.com
```

rootmelo92118 (changes requested):
Please sort it.

PR #2189, opened 2024-06-13
Add Longbridge securities

xkww3n:


rootmelo92118 (changes requested):
Please include it into relevant category.

xkww3n:
> Please include it into relevant category.

I've included it in category-finance.
It's a Hong Kong based company so it can't be included in category-securities-cn, and it has CDN nodes in Mainland China so also can't be included in geolocation-!cn.

rootmelo92118 (approved):

PR #2192, opened 2024-06-13
Update jetbrains-ai

De54:
It is better to define these subdomains as `domain:` because I have encountered IP leaks

One of the sniffed subdomains in wireshark:

Subdomain: `api.app.prod.grazie.aws.intellij.net`

Screenshot_20240613_205133-1


de54:
@rootmelo92118 why?

PR #2206, opened 2024-06-17
Add `!cn` tag for some ctrip domains.

xkww3n:


xkww3n:
Merged into #2207, closing

PR #2211, opened 2024-06-21
Update go.mod

rootmelo92118:
It just a new try.
If it has any problem after merge, it will be revert immediately.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/v2fly/domain-list-community
 
-go 1.21
+go 1.22
 
 toolchain go1.22.2
 


IceCodeNew:
Better not to touch it unless you are introducing a dependency that require higher minimum version of golang. The version here is basically indicating the minimum version this project supports.
Anyway, since it seems all good after bumping that version, no action of reverting is required.

I'd like to reassure you that you can still have someone review your PRs. While I may not be as actively involved with many of the PRs here due to shifting interests, I remain committed to and concerned about this project."

IceCodeNew:
Better not to touch it unless you are introducing a dependency that require higher minimum version of golang. The version here is basically indicating the minimum version this project supports.
Anyway, since it seems all good after bumping that version, no action of reverting is required.

I'd like to reassure you that you can still have someone review your PRs. While I may not be as actively involved with many of the PRs here due to shifting interests, I remain committed to and concerned about this project.

PR #2226, opened 2024-07-06
Add kucoin.com

closehandle:


data/kucoin
```diff
@@ -0,0 +1,3 @@
+kucoin.com
+kucoin.plus
+staticimg.com
```
rootmelo92118, line 3: Please provide evidence of this domain belonging to this category.

rootmelo92118 (changes requested):

data/kucoin
rootmelo92118: No. I mean the `staticimg.com`

closehandle: KuCoin hosts static resources on it. I searched through all the rules and didn't find any other references to `staticimg.com`, so I added it in

rootmelo92118: > 
> 
> KuCoin hosts static resources on it. I searched through all the rules and didn't find any other references to `staticimg.com`, so I added it in

No, I mean... Are you sure this domain is only use on this website?

rootmelo92118:
According the searching result in google, I am very sure this doamin is not just use on this website.
image

PR #2239, opened 2024-07-12
Add Russian TLDs to `category-gov-ru`

demarcush:
Since there's no official `category-ru`.


rootmelo92118:
I don't think this is a good idea because the TLD `ru` is not just use for russian government.

PR #2249, opened 2024-07-18
Improved collection of category-acg

xkww3n:


rootmelo92118:
Actually, I think this PR does not have any sense. Do you have any opinion?

xkww3n:
> Actually, I think this PR does not have any sense. Do you have any opinion?

Previously this file contains almost nothing, this PR adds some ACG-related companies, games, etc. to it to make it fits to its name.

data/category-acg
```diff
@@ -1,5 +1,30 @@
+include:abema
+include:acfun @cn
```
IceCodeNew, line 2: 1. a category-acg-cn would be better
2. corp. like acfun may also have some sites mainly serving the oversea users. Labelling a sub-category does not practically make sense, see the point 1

data/geolocation-!cn
```diff
@@ -1,8 +1,5 @@
 # This list contains domains that don't have access point in China mainland. This is opposite to geolocation-cn.
 
-# Anime, Comics & Games
-include:category-acg
```
IceCodeNew, line 4: Not a good idea.
You deleted domains from the collection `geolocation-!cn` for no good reasons.

IceCodeNew (changes requested):

data/category-acg
xkww3n: Then how about have it split to category-acg-cn and category-acg-!cn?

PR #2260, opened 2024-07-22
Create ntt and trackernetwork

User53464563:


data/geolocation-!cn
```diff
@@ -129,6 +130,7 @@ include:streamable
 # Forums & Social network
 include:category-forums
 include:category-social-media-!cn
+include:category-communication
```
IceCodeNew, line 133: I did not get it.

data/geolocation-!cn
```diff
@@ -58,8 +58,9 @@ include:stackpath
 # Certificate authorities
 include:category-cas
 
-# Communications
-include:category-communication
+# Communication
+include:ntt
+include:softbank
```
IceCodeNew, line 63: Not all sites of softbank are related to communication, I guess

IceCodeNew (changes requested):

data/geolocation-!cn
User53464563: I tend to move "category-communication" items into "category-social-media-!cn", but that might be a breaking change

data/geolocation-!cn
User53464563: i reverted this change

PR #2276, opened 2024-08-01
Create category-bank-jp

ImAubrey:


data/category-bank-jp
rootmelo92118, line 1: Please include it to relevant category.

rootmelo92118 (changes requested):

data/category-bank-jp
IceCodeNew: That is fine.

IceCodeNew (approved):

PR #2291, opened 2024-08-08
feat: add fastly cdn address in steam@cn

JxQg:
This cdn link is for downloading in cn.

### DIFF
--- a/data/steam
+++ b/data/steam
@@ -48,6 +48,8 @@ full:steamusercontent-a.akamaihd.net
 full:steamuserimages-a.akamaihd.net
 full:steamvideo-a.akamaihd.net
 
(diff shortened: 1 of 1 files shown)


IceCodeNew:
> This cdn link is for downloading in cn.

Any resource mentioned this domain?

JxQg:
I think when I download games in the China-AnyCity region, my Clash core uses this CDN domain. However, I have only tested that it can connect directly using the CMCC network, and I don't have the ability to test it on other networks.

JxQg:
@IceCodeNew Sry. It's use China-HK region download maybe. I can't recurrence this. I will close this PR.

PR #2299, opened 2024-08-14
Add several academic domains

ysakura99:


IceCodeNew (changes requested):
1. sort
2. add annotations

PR #2330, opened 2024-09-08
Proposal to add `category-ntp` and `googlefcm` to `private`

demarcush:


rootmelo92118 (changes requested):
Could you give me a reason to do this operation?

database64128:
I don't think this is the intended use of the private list. If you don't want to use your proxy for destinations in these lists, just add routing rules for them.

PR #2334, opened 2024-09-12
Split private domain list

adventureRyan:
Currently the private list consists of there different kinds of domains:
- **Domains reserved for local network**, for example 'miwifi.com' and domains defined by RFC6762.
  These domains should be resolved locally or are resolved to local addresses and usually do not work well with fake-ip.
- **Domains defined by OS or applications for connectivity checking purpose**, such as 'msftconnecttest.com'.
  These domains are fine with fake-ip since captive portal detections are implemented with HTTP requests and domains are resolved to public addresses.
- Others like 'push.services.mozilla.com'.
  The reason of being added remains unknown.

This PR splits domains into different lists:
- Reserved local domains remain untouched.
- Connectivity checking domains are moved into separate list, connectivity-check.
- Others are moved back to service providers' list for now.

And appends Debian connectivity check service to connectivity-check list.
(ref: https://salsa.debian.org/utopia-team/network-manager/-/blob/b944168515fa1bef3fcd80895b2122f003605a08/debian/20-connectivity-debian.conf#L2)

### DIFF
--- a/data/connectivity-check
+++ b/data/connectivity-check
@@ -1,0 +1,27 @@
+# ArchLinux
+ping.archlinux.org
+
+# Apple
+captive.apple.com
+
+# Android
+connectivitycheck.gstatic.com
+
+# Cloudflare
+cp.cloudflare.com
+
+# Debian
+network-test.debian.org
+
+# Firefox
+detectportal.firefox.com
+
+# KDE
+networkcheck.kde.org
+
+# MIUI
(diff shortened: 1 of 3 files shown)


database64128:
> And then once the url of it is bypassed, you get to receive your notifications even when your proxy service is not working properly.

What does this mean?

database64128 (approved):
LGTM.

PR #2355, opened 2024-09-25
Add `signalr.net` to `microsoft`

posthumz:
ref: https://signalr.net/


posthumz:
`signalrs2-relayhub-prod-as01-1.service.signlalr.net` is used by `CrossDeviceService` and `PhoneExperienceHost` in Windows, and is available in mainland China.

PR #2371, opened 2024-10-11
Adding Supercell and separating Google search

molink36:
Google search domains are separated from other Google domains.
Supercell domains are added.


data/google
rootmelo92118, line 1: The domains that you removed are not only for search. They already have a lot of sub domains on them.

rootmelo92118 (changes requested):

molink36:
Dear @rootmelo92118

Thank you. You are right.
Changes are applied.

rootmelo92118 (approved):

PR #2372, opened 2024-10-13
Adding Russian domains blocked outside

eldevel:


data/category-ru-blocked-outside
rootmelo92118, line 1: This category cannot be merged in this project. We cannot accept another GFW list that exists here.

rootmelo92118 (changes requested):

PR #2437, opened 2024-12-15
feat: Added ru blocked domains from ru and from worlds for ru users.

Jolymmiles:
zapret - https://antifilter.download/list/domains.lst
zapret-zapad - https://community.antifilter.download/list/domains.lst


Jolymmiles:
@Loyalsoldier

PR #2459, opened 2024-12-28
Add cryptomus

closehandle:
https://cryptomus.com


rootmelo92118 (changes requested):
If it only just has one domain, you don't need to create a new category.

data/category-cryptocurrency
```diff
@@ -6,6 +6,7 @@ include:bitflyer
 include:bitsquare
 include:bybit
 include:coinone
+include:cryptomus
```
rootmelo92118, line 9: Remove it.

closehandle: Oh no, i forget about it, now removed.

rootmelo92118 (approved):

PR #2475, opened 2025-01-10
Comfy

zfkun:
- add comfy
- add category-ai-drawing-!cn

### DIFF
--- a/data/category-ai-drawing-!cn
+++ b/data/category-ai-drawing-!cn
@@ -1,0 +1,4 @@
+# AIGC drawing‌ related
+
+# Stable Diffusion
+include:comfy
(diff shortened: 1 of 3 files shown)


EpLiar:
Why do domain data exclude `comfy.org`?

zfkun:
> Why do domain data exclude `comfy.org`?

The main domain `comfy.org` is normally accessible in the China.
So it wasn't included.

data/category-ai-drawing-!cn
rootmelo92118, line 1: This category should be included in the others category.

rootmelo92118 (changes requested):

MkQtS:
`comfy.org` resolves to AWS anycast. It's better to add it.

PR #2505, opened 2025-01-31
Create claude

Aethersailor:


PhoenixEmik (changes requested):
Duplicated with anthropic.

Aethersailor:
> There's already an anthropic list.
> 
> You may change it to `include:anthropic` instead.

thanks

PR #2524, opened 2025-02-13
Multipple domains are added or updated

molink36:


rootmelo92118 (approved):

data/category-anticensorship
```diff
@@ -27,13 +28,16 @@ getoutline.org
 gfw.report
 guardianproject.info
 hiddify.com
+ipleak.com
+ipleak.net
```
rootmelo92118, line 32: Lines 21, 31, and 32 are not appropriate for inclusion in the anti-censorship category because they merely serve as websites to test your network environment, not to circumvent Internet censorship.

rootmelo92118 (changes requested):

data/category-anticensorship
molink36: I removed Lines 21, 31, and 32 ("dnsleaktest.com", "ipleak.com" and "ipleak.net"). However, in that regard, line 16 ("browserleaks.com") is the same, and I believe it should also be removed. I think a category for these kinds of websites should be created.

IceCodeNew:
After @rootmelo92118 finish reviewing, pls let me know. I would like to be envolved and to do the final decision on this PR.

rootmelo92118:
> After @rootmelo92118 finish reviewing, pls let me know. I would like to be envolved and to do the final decision on this PR.

I am done.

IceCodeNew (approved):

PR #2566, opened 2025-03-12
Add assets.adobedtm.com botframework.com

written143:


data/xbox
```diff
@@ -4,6 +4,7 @@ include:mojang
 # All .xbox domains
 xbox
 
+adobedtm.com
```
rootmelo92118, line 7: This domain is belonging to `adobe` not `xbox`

rootmelo92118 (changes requested):

rootmelo92118 (approved):

PR #2583, opened 2025-03-27
Update category-ecommerce

Aethersailor:
Shopify and Ozon are both well-known e-commerce websites, and they already have independent rules.

Add "include" rules for Ozon and Shopify in category-ecommerce.

### DIFF
--- a/data/category-ecommerce
+++ b/data/category-ecommerce
@@ -22,6 +22,7 @@ include:pchome
 include:rakuten
 include:ruten
 include:shopee
+include:spotify
 include:suruga-ya
 include:target
 include:walmart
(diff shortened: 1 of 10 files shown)


data/category-ecommerce
```diff
@@ -17,10 +17,12 @@ include:momo
 include:muji
 include:nike
 include:nordstrom
+include:ozon
 include:pchome
 include:rakuten
 include:ruten
 include:shopee
+include:spotify
```
IceCodeNew, line 25: this is not the shopify

IceCodeNew (changes requested):

data/category-ecommerce
Aethersailor: Sorry, I entered it incorrectly before, and it has been corrected.

Aethersailor: > this is not the shopify

Sorry, I entered it incorrectly before, and it has been corrected.

IceCodeNew (approved):

PR #2606, opened 2025-04-06
add google domain

yueisme:
`app-analytics-services.com`: https://github.com/firebase/firebase-ios-sdk/issues?q=app-analytics-services.com

### DIFF
--- a/data/google-ads
+++ b/data/google-ads
@@ -43,5 +43,6 @@ marketingplatform.google.com @ads
 urchin.com @ads
 
 full:analytics.google.com @ads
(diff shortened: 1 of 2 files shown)


data/google-ads
```diff
@@ -43,5 +43,6 @@ marketingplatform.google.com @ads
 urchin.com @ads
 
 full:analytics.google.com @ads
+full:ogads-pa.clients6.google.com @ads
```
IceCodeNew, line 46: make another pr and provide background info

IceCodeNew (changes requested):

PR #2624, opened 2025-04-15
Create category-game-platforms-download

Aethersailor:
Create a category that contains the download CDN domains of many well-known game platforms, such as Steam, Epic Games, Battle.net, and EA Desktop, etc.

***

Users can use this category and `category-game` at the same time, and place `category-game-platforms-download` before `category-game`.

In this way, users can achieve the following functions: `PROXY` access to the game platform and `DIRECT` access to the game download CDN at the same time.

***

The domains in category are collected from the hosts tool `UsbEAm Hosts Editor V4.0.1`.

Maybe there are more game platform download CDN domains that are not included. 

I will update this category regularly and add as many game platform download CDN domains as possible.  

***

PS:
I did not use `regexp:` but `full:` for the following reasons:
1. To match CDN domains more accurately
2. Other contributors can add or delete domains more easily in the future


IceCodeNew (approved):

IceCodeNew:
> I did not use `regexp`

that is encouraged ;-)

PR #2636, opened 2025-04-19
Update razer

pover0k:


data/razer
```diff
@@ -7,3 +7,8 @@ razerzone.com @cn
 razerzone.jp
 respawnbyrazer.com
 rzr.to
+
+full:b6y6lf1s29.execute-api.us-east-1.amazonaws.com
+
+# tracking
+full:o72136.ingest.sentry.io @ads
```
rootmelo92118, line 14: Maybe you can move this domain to `razer-ads` and include it in this category, and `category-ads`.

rootmelo92118 (changes requested):

data/razer
```diff
@@ -7,3 +7,8 @@ razerzone.com @cn
 razerzone.jp
 respawnbyrazer.com
 rzr.to
+
+full:b6y6lf1s29.execute-api.us-east-1.amazonaws.com
```
IceCodeNew, line 11: any background info?

IceCodeNew: that query may be triggered by another application.

data/razer
```diff
@@ -7,3 +7,6 @@ razerzone.com @cn
 razerzone.jp
 respawnbyrazer.com
 rzr.to
+
+# tracking
+full:o72136.ingest.sentry.io @ads
```
IceCodeNew, line 12: more background info needed

PR #2666, opened 2025-04-30
ctrip: add more domains

MkQtS:


yk271:
I think we need to give more consideration to the domain tripcdn.com. It's also used by trip.com, Ctrip's international site. Some of its subdomains, such as aw-s.tripcdn.com and ak-s.tripcdn.com, do not have CDN nodes in China.

Since the ctrip rules are included in the geolocation-cn rules, this can lead to extremely slow response times for these specific subdomains when accessed directly from mainland China.

I suggest that we add the @!cn tag to these affected subdomains to address this issue.

PR #2678, opened 2025-05-07
Update geolocation-cn

NextAbyss:
Add some chinese domain for wechat miniapp.
Add Xiaoheihe app domain.

### DIFF
--- a/data/category-pt
+++ b/data/category-pt
@@ -74,7 +74,11 @@ m-team.io @!cn
 
 myanonamouse.net @!cn
 torrentleech.org @!cn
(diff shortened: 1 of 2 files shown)


IceCodeNew:
Should put these domains in a seperate list
whay do you think? @rootmelo92118

IceCodeNew:
After @rootmelo92118 finish reviewing, pls let me know. I would like to be envolved and to do the final decision on this PR.

rootmelo92118:
> Should put these domains in a seperate list whay do you think? @rootmelo92118

I agree with your opinion.

PR #2689, opened 2025-05-18
Create pornolab

bruneivan:
Added 'pornolab.net' domain (torrents)

### DIFF
--- a/data/pornolab
+++ b/data/pornolab
@@ -1,0 +1,1 @@
+pornolab.net


rootmelo92118 (changes requested):
Please include it in the relevant category.

MkQtS:
Already in `category-porn`

https://github.com/v2fly/domain-list-community/blob/db926488c8232d6e4f4347546b057364bcb66607/data/category-porn#L4092

There seems no much effect to create such a single entry.

PR #2720, opened 2025-05-26
Increase Performance 

Pk-web6936:


IceCodeNew:
I cannot see how this PR can increase performance. You may re-open this PR with some benchmark
P.S. We currently do not care about performance.

PR #2755, opened 2025-06-18
Adding Google Labs domains

vemneyy:
Domains has been added to the list of domains to support Google Labs services. This change is necessary for the functionality interacting with Google's experimental AI services hosted on this domain.


vemneyy:
@Loyalsoldier @rootmelo92118 @IceCodeNew

PR #2767, opened 2025-06-30
Added Yandex ads domains

vkc-code:
Added list of Yandex ads domains (Yandex is kind of Russian Google)


Loyalsoldier:
Advertising or tracking domains are hard to maintain, it's not encourage to add more. Thanks though.

PR #2769, opened 2025-07-01
Update Nvidia domains with CDN subdomains

easymoney322:
I've noticed there are connections from Nvidia APP to the following domains, which are CNAMES:
download.gfe.nvidia.com.edgekey.net
ngx.download.nvidia.com.edgekey.net
services.gfe.nvidia.com -> services.gfe.nvidia.com.edgesuite.net
static.nvidiagrid.net -> static.nvidiagrid.net.edgekey.net
ops.gx.nvidia.com.edgesuite.net
origin-gfwsl.gtm.nvidia.net -> gfwsl.geforce.com.edgesuit.net
ota.nvidia.com -> ota.nvidia.com.edgekey.net
www.nvidia.com -> www.nvidia.com.edgekey.net

I believe these are subdomains belonging to nvidia on Microsoft's CDN

### DIFF
--- a/data/nvidia
+++ b/data/nvidia
@@ -61,6 +61,7 @@ tegrazone.jp
 tegrazone.kr
 
 geforce.com.edgesuite.net
+nvidia.com.edgesuite.net
 nvidia.com.edgekey.net
 nvidiagrid.net.edgekey.net
 full:nvidia.tt.omtrdc.net
--- a/data/nvidia
+++ b/data/nvidia
@@ -43,6 +43,7 @@ nvidia.it
 nvidia.jp
 nvidia.lu
 nvidia.mx
+nvidia.net
 nvidia.nl
 nvidia.no
 nvidia.pl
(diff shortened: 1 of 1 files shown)


Seikilos:
This is confusing. Edgekey and edgesuite is an Akamai CDN.
Nvidia.net does not appear to be directly owned by nvidia, but by safenames.net.

IMHO I would skip nvidia.net and I would add edgekey and edgesuite to AKAMAI instead

easymoney322:
Hello,
It is confusing, yes, I saw `spotify.com` using something simmilar, but with `full` domain.
There is, however, a lot of fqdns in nvidia's subdomain in CDN, so I thought redirecting whole subdomain would make sense.

How should I change it? I want to redirect all nvidia's traffic (including traffic to the CDN) via this file, but don't route the CDNs itself.

Seikilos:
@easymoney322 I edited my reply and changes "I would commit nvidia.net" to "I would skip nvidia.net". The ownership is unclear, so I personally would not want to have this there.
On your question: I cannot answer it, I am not a maintainer and don't really know the methodology here, sorry. 
This is tricky to solve since CDN urls may even be designed to be highly dynamic and ownership on those CDN deep links may randomly change, too.

Loyalsoldier:
Actually, it's unnecessary to add CNAME domains.

PR #2787, opened 2025-07-18
Update tld-cn

universonic:
依据[互联网号码分配机构 (IANA)](https://www.iana.org/domains/root/db)最新的根域列表更新


Loyalsoldier:
Updated recently, see PR https://github.com/v2fly/domain-list-community/pull/2816

PR #2795, opened 2025-07-25
Add more tapsi-related domains to category-travel-ir

sadeghi-aa:
Apart from tapsi's main domain which is tapsi.ir, if you visit the PWA version of the app hosted on https://app.tapsi.cab, you could see that some of the resources are served from other domains. These must also be exempted from v2ray so that tapsi works fine while v2ray is connected.

### DIFF
--- a/data/category-travel-ir
+++ b/data/category-travel-ir
@@ -4,3 +4,6 @@ alibaba.ir
 flightio.com
 mrbilit.com
 tapsi.ir
+tapsi.cab
+tap33.me
+tap30.services


MkQtS:
I did a search on GitHub, and found these domains in some Iraian repos. So the merge should be OK.

Due to the network restrictions and lacking of manpower, it's hard for us to view such changes. You may consider using/contributing to other relevant repos, e.g. https://github.com/bootmortis/iran-hosted-domains. (You are still welcome to pr here)

sadeghi-aa:
@MkQtS  Thank you for mergin the PR. I got a question. Do you periodically check those repos to update this repo? I mean, does contributing to those repos actually have an effect on v2ray clients?

MkQtS:
Contributing to other repos will not affect this repo. I mean you can use geodata file from other (Iraian optimized) sources.

In order to use geodata from other sources, you need to download the desired `geosite.dat` and put it to the corect path. Some clients like mihomo, v2rayN may support to specify the geodata source explicitly.

PR #2826, opened 2025-08-25
Add papers.cool in category-scholar-cn

NeumoNeumo:
`papers.cool` is a paper overview tool. Accessing from a foreign country is much slow than visiting it from China directly.


NeumoNeumo:
Could you please let me know the reason this pull request was closed? Should I add it in `category-ai-cn`?

Loyalsoldier:
The resolved IP address of `papers.cool` belongs to `HK`, so it should be proxied by default in this project.

PR #2854, opened 2025-09-05
Create aliexpress

ingria:
sniffed from the ru website and ios app


MkQtS:
Aliexpress is running by alibaba. Most domains here were already added in `data/alibaba`. And it's not likely to seperate it. Please use `geosite:alibaba` instead.

PR #2886, opened 2025-10-09
Add category DNS-over-HTTPS

Meo597:
Sometimes, for certain purposes, you might want to block them

### DIFF
--- a/data/category-doh
+++ b/data/category-doh
@@ -1,3 +1,5 @@
+# This is not related with `category-httpdns-cn`
+
 # **Public anycast resolvers**
 
 # AdGuard DNS


Meo597:
Data source: https://raw.githubusercontent.com/AdguardTeam/KnowledgeBaseDNS/refs/heads/master/docs/general/dns-providers.md

Simply cleaned with regular expressions and then manually reviewed, there may be duplicates.

MkQtS:
1. Rename to category-encrypted-dns since some of them use DoT not DoH
2. Avoid duplicates (`dns.bebasid.com`, `doh.cleanbrowsing.org`, `freedns.controld.com`, `internetsehat.bebasid.com`)
3. Avoid unnecessary subdomains (subdomains of `cloudflare-dns.com`, `dns.mullvad.net`, `dns.nextdns.io`, `dns.sb`, `dns.seia.io`, `doh.pub`, `doh.sb`, `futuredns.eu.org`). Remove/Comment out them or use `full:` for all domains
4. Include `category-httpdns-cn` at line 1 but comment it out (as `category-httpdns-cn` are using unstandard protocol or used in private apps)

Meo597:
1. All domains in this list are DoH. Some providers may also be running DoT services on the same domain, but these should not be considered.
2. Duplicates have been processed.
3. All domains changed to `full:`.
4. This has nothing to do with httpdns-cn. I don't understand your intention.

MkQtS:
> Some providers may also be running DoT services on the same domain, but these should not be considered.

> This has nothing to do with httpdns-cn. I don't understand your intention.

DoT, DoH and httpdns are different protocols. But to some extent, their purposes are consistent: bypassing plain DNS. Some end users may be confused at these different entries.

Considering you are adding a new entry for DoH servers, there may be another request for DoT servers in the future. Since many DoH and DoT services are provoded by the same companies and even using the same domains, I think it's better to use a single `category-encrypted-dns` rather than two entries with some duplicate content and similar usage scenarios.

Meo597:
If the end user wants to handle DoT, they only need to simply use TCP UDP 853 as a condition.
There are almost no side effects.

This list was collected from AdGuard and then semi-automatically batch processed.
Independent DoT has been excluded

If you insist on renaming it to `category-encrypted-dns`, that's fine.
However, the DoT part will need to be added gradually by other maintainers

MkQtS:
算了，就这样吧，如果有人提再说

PR #2906, opened 2025-10-22
Create x5

vladon:
X5 is a is Russia's largest food retailer.  Some people may need to use this domain collection.


MkQtS:
1. Any proof about why these domains belong to x5?
2. Sort these domains
3. Include it in `category-ru`

vladon:
whois of each domain? I just work there (in X5)

fix: sorted, added to category-ru

PR #2910, opened 2025-10-26
Add nodeseek

kagangtuya-star:
This is a forum for discussing servers.


MkQtS:
You don't need to add subdomains if the main domain is added.

Please remove `www.*.com` add include nodeseek in `category-forums`

PR #2933, opened 2025-11-07
Split bilibili-cdn from bilibili

H1JK:
`bilibili-cdn` contains CDN domains that host images, videos, and static assets for bilibili that are irrelevent to business logic.

All domains are extracted from the existing `bilibili` category; no new domains have been added. `bilibili` now contains `bilibili-cdn` and should not be changed by this.

Usage: when you use proxies (just want) to bypass regional restrictions, you can securely route CDN domains to "direct outbound", thus saving proxy data traffic.

### DIFF
--- a/data/bilibili
+++ b/data/bilibili
@@ -1,3 +1,4 @@
+include:bilibili-cdn
 include:bilibili-game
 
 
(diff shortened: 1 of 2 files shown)


H1JK:
Off topic: idk what `im9.com` and `yo9.com` are used for, so I can't determine if they are CDN domains. I just kept them in `bilibili`.

PR #2949, opened 2025-11-12
Add 2ch.org and 2ch.su to 2ch

krisavdome:
2ch.life is dead, 2ch.org and 2ch.su are the actual domains as for now, 2ch.hk redirects to 2ch.org

### DIFF
--- a/data/2ch
+++ b/data/2ch
@@ -1,3 +1,4 @@
 2ch.hk
+2ch.life
 2ch.org
 2ch.su


MkQtS:
Are you sure that `2ch.life` is dead? I can still visit `https://2ch.life/wh/`

krisavdome:
oh, my bad, couldn't access it bc it got blocked. i thought that they disable old domains

PR #2989, opened 2025-11-28
Add ofiii domains

NNdroid:


MkQtS:
Both `ofiii` and `litv` belongs to TGC Taiwan Inc. `ofiii` also uses `svc.litv.tv`.

There is already `data/litv` in this repo, so you can just add new domains(`ofiii.com` and `full:ntdofifreepc-tgc.cdn.hinet.net`) to `data/litv`.

NNdroid:
Okay, I'll modify the code.

PR #2990, opened 2025-11-28
category-entertainment: deduplicate

MkQtS:


iLeonidze:
@MkQtS, perhaps it would have been better to remove the domain `thetvdb.com` from the entertainment category instead of deleting the entire `tvdb` group? Right now everyone who uses the tvdb group is getting an error.

MkQtS:
@iLeonidze fixed now. Sorry for the inconvenience.

PR #3005, opened 2025-12-01
Edit mailru-group

Pavel-Boyazov:
Some edits with mailru, ok, vk, mailru-group and add dzen


MkQtS:
I modified your commits and force-pushed. ~~Please check for any omissions.~~

If there are any omissions, feel free to open another PR.

PR #3017, opened 2025-12-07
Add qobuz

ninenfn:


MkQtS:
Subdomains are unnecessary. Check the README for details.

Unless there is a specific reason, please do not add a new entry that contains only one domain. And a new entry should be ideally included in a relevant category(`category-entertainment`).

My suggestion is just adding `qobuz.com` in `category-entertainment`.

PR #3022, opened 2025-12-08
Add 'r2.dev' to Cloudflare data

malagebidi:
Cloudflare-managed https://r2.dev subdomain for public buckets


MkQtS:
`https://developers.cloudflare.com/r2/buckets/public-buckets/`

PR #3039, opened 2025-12-11
Add more ad domains

DeepChirp:
参见：
* easylist/easylist#10219


MkQtS:
I suggest not adding `stats.hanmaker.com` as it's not widely used and this repo is not designed for adblocking.

DeepChirp:
> I suggest not adding `stats.hanmaker.com` as it's not widely used and this repo is not designed for adblocking.

Okay, it has been removed. Perhaps later I should only add advertising domains that are widely used, or corresponding to existing company lists.

PR #3053, opened 2025-12-15
Add Ookla Speedtest Central to the list

tooadstool:


Loyalsoldier:
I think it makes no sense to just include a third-party speedtest server.

PR #3056, opened 2025-12-16
Fix #3055

### LINKED ISSUE #3055: [bug] geosite:geolocation-!cn includes look ahead regexp which cause dae crash
```
# Image Resource Domain like `img.diwodiwo.xyz` `s3.diwodiwo.xyz` `storage.diwodiwo.xyz` `storage-b.diwodiwo.xyz`
regexp:^([a-z0-9-]+\.)*(?!ad-display\.|ad-channel\.)[a-z0-9-]+\.diwodiwo\.xyz$
```

这段会导致某些软件（如dae）报错，希望能 revert。  或者加到单独的文件里，不要加到 geosite:geolocation-!cn 这个被广泛使用的分类里。

初步判断应该是 (?! 这类复杂的 look ahead 正则导致的问题，建议用其他方式处理，避免用 这类正则。

PR which cause this issue: https://github.com/v2fly/domain-list-community/pull/2695

luochen1990:
Accroding to README.md:

> Note: Adding new regexp and keyword rules is discouraged because it is easy to use them incorrectly, and proxy software cannot efficiently match these types of rules.

### DIFF
--- a/data/picacg
+++ b/data/picacg
@@ -10,6 +10,8 @@ picacomic.xyz
 wikawika.xyz
 
 # Image Resource Domain like `img.diwodiwo.xyz` `s3.diwodiwo.xyz` `storage.diwodiwo.xyz` `storage-b.diwodiwo.xyz`
-regexp:^([a-z0-9-]+\.)*(?!ad-display\.|ad-channel\.)[a-z0-9-]+\.diwodiwo\.xyz$
+
 ad-channel.diwodiwo.xyz @ads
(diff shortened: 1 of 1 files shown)


Meo597:
Unless absolutely necessary, avoid using regular expressions.
They are not only prone to errors (this is an example).
But also have terrible performance. Domain rules don't actually have such strict requirements.
`domain` and `keyword` are both fast

PR #3057, opened 2025-12-16
add syntax check

AzunyanDaisuki:


Loyalsoldier:
The same check done in https://github.com/v2fly/domain-list-community/pull/3064 . Thanks though.

PR #3058, opened 2025-12-16
Add wink.ru streaming domains, update category-ru

sklimoff:
wink.ru - russian video-streaming service. Add it main doman and technical streaming to list.


MkQtS:
Please sort the include and domains alphabetically

PR #3062, opened 2025-12-16
Add a check for bad regexp

Gaojianli:
Hello guys, I add a test for the regexp, in order to avoid #2695 happens again


luochen1990:
I think that's a good idea. why you close it, what problem you got, do you need any help?

Gaojianli:
> I think that's a good idea. why you close it, what problem you got, do you need any help?

The source branch is wrong, so I recreated one: https://github.com/v2fly/domain-list-community/pull/3063

PR #3063, opened 2025-12-16
Add a check for bad regexp

Gaojianli:
Hello guys, I add a test for the regexp, in order to avoid #2695 happens again.

Here's a example of bad pr: https://github.com/Gaojianli/domain-list-community/pull/1


Loyalsoldier:
The same check done in https://github.com/v2fly/domain-list-community/pull/3064 . Thanks though.

PR #3079, opened 2025-12-19
Add `ir` category as an alias of `category-ir`

patterniha:
Almost no one would guess that Iranian domains are in ‍‍‍‍`category-ir` unless they look at this repo!

Since Iranian IPs are in `geoip:ir`, Iranian domains must also be in `geosite:ir`

### DIFF
--- a/data/ir
+++ b/data/ir
@@ -1,0 +1,2 @@
+# Same as "category-ir"
+include:category-ir


patterniha:
?

krisavdome:
but its logically a category, all of the country based domain lists are categories, renaming it will break existing configs

patterniha:
> but its logically a category, all of the country based domain lists are categories, renaming it will break existing configs

This is not rename, this adds `ir` category and it will be equal to `category-ir`. (it is just an alias)

we have `geosite:cn` for Chinese domains, so we should have same names for other countries.
however, for compatibility `category-ir` remains unchanged.

krisavdome:
i missed that part somehow lol

MkQtS:
GeoIP uses standardized two-letter abbreviations for countries, but that's not the case for geosite. Geosite uses another naming conventions. We have not added such aliases for countries before. (`data/cn` is there for compatibility, and is not completely alias)

I don't think this change is necessary. For users intending to write their own geosite rules, they need to learn which `geosite:entry` are available.

PR #2918, opened 2025-10-30
Immersivetranslate.com has no access points in mainland

Zeehan2005:
```
>nslookup immersivetranslate.com 114.114.114.114
服务器:  UnKnown
Address:  114.114.114.114

名称:    immersivetranslate.com
Addresses:  172.66.149.158
          104.20.23.243
```

[172.66.149.158 United States/California/San Francisco](https://whatismyipaddress.com/ip/172.66.149.158)
[104.20.23.243 United States/California/San Francisco](https://whatismyipaddress.com/ip/104.20.23.243)

Even though it has a mainland ICP number, it doesn't has access points in mainland.
虽然有大陆ICP备案，但在大陆没有接入点。

PR #3083, opened 2025-12-19
Immersivetranslate

Zeehan2005:
Continuation of #2918 

Websites with the suffix "cn" are operated by 上海书同文网络科技有限公司, a Chinese company, and have mainland access points and ICP 备案. Websites with the suffix "com" are the international versions and do not have mainland access points.

@Aethersailor


Aethersailor:
感谢修复

PR #3087, opened 2025-12-20
Remove 'habr' from category-ru

hzhexee:
Removed 'include:habr' from the category-ru file.

Reasons:

1) habr.com isn't affiliated with Russia directly (even though the habr.com IS hosted in Russia, it's cdn domain - habrastorage.org is hosted in Estonia) 
2) the site itself contains a lot of geoblocked articles which can't be accessed without a foreign proxy - so it being in category-ru can lead to routing mishaps in cases such as routing category-ru sites to DIRECT on the client

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,11 @@ on:
       - master
     paths-ignore:
       - "**/README.md"
+
+concurrency:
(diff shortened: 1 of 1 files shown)


MkQtS:
Thanks for the clarification.

PR #3115, opened 2025-12-30
Update google-ads

bafometik:
adtrafficquality.google
fcmatch.google.com
analytics-alv.google.com
afs.googleusercontent.com
ogads-pa.clients6.google.com


MkQtS:
Why do you think `afs.googleusercontent.com` and `ogads-pa.clients6.google.com` are related to ads? What's your basis?

bafometik:
i saw them in adblock lists, maybe i shouldn't have to add them, idk

PR #3117, opened 2025-12-31
category-ru: add anime domains

6kembion9-byte:


MkQtS:
The file name makes no sense and may mess things up. You can add these two domains in `data/category-acg` or `data/category-ru`.

PR #3140, opened 2026-01-06
category-cryptocurrency: add bingx.com

ghost:
## create category-NAME to include all category-NAME-*

Expected tree like this:

```
category-NAME (contains all child)
    \---- category-NAME-cn
    \---- category-NAME-!cn
    \---- category-NAME-ir
    \---- category-NAME-fr
    \---- category-NAME-it
    \---- category-NAME-ko
    \---- category-NAME-jp
    \---- category-NAME-ru
    \---- category-NAME-dk
```
(Also I want to rename `category-ads-all` to `category-ads` because of above tree rule. What do you guys think?)

### DIFF
--- a/data/category-cryptocurrency
+++ b/data/category-cryptocurrency
@@ -25,6 +25,7 @@ include:zb
 
 1inch.io
 aave.com
+bingx.com
 bitbank.cc
 bitcoin.org
 bitfinex.com


MkQtS:
I don't think such new categories serve any practical purpose.

PR #3170, opened 2026-01-12
Update releases compression

Oliveoilfly:
- (non-breaking, visual only) Combine xz POSIX-style options in one
- (non-breaking) Add fast&efficient zstd compression


Loyalsoldier:
Compressed assets are just for backward compatibility, it's basically useless.

PR #3175, opened 2026-01-13
Add pubg

Zeehan2005:
PUBG related

### DIFF
--- a/data/pubg
+++ b/data/pubg
@@ -1,4 +1,5 @@
+kraftonde.com
 playbattlegrounds.com
 pubg.com
-kraftonde.com
-pubg1.battleye.com
+
+full:pubg1.battleye.com


MkQtS:
Regexp is discouraged. It seems that only `pubg1.battleye.com` resolves to a valid IP address.

There are `pubghelper.com`, `pubgno1.cn` and `pubgtool.com` in `data/tencent-games` now. Please check if they should be moved to the new `data/pubg`

Zeehan2005:
> Regexp is discouraged. It seems that only `pubg1.battleye.com` resolves to a valid IP address.
> 
> There are `pubghelper.com`, `pubgno1.cn` and `pubgtool.com` in `data/tencent-games` now. Please check if they should be moved to the new `data/pubg`

The domain name in this file is used by _PUBG_ (PC game) operated by KRAFTON, and has nothing to do with Tencent. Tencent seems to have cooperated with KRAFTON to release many mobile games both within and outside mainland China, such as _PUBG Mobile_, _刺激战场_, and _全军出击_. Moving may cause incorrect traffic diversion, leading to access issues. 

For Regexp, I have replaced it. Thanks for your information.

MkQtS (approved):

PR #3192, opened 2026-01-19
Add liveperson and standardchartered

futuretech6:


MkQtS:
1. Does LivePerson belong to HSBC? If not, then it should not add to hsbc
2. Why do you name it `schk` instead of `standardchartered`? It seems to be a UK company
3. Please change `online.standardchartered.com` to `standardchartered.com`

futuretech6:
> 1. Does LivePerson belong to HSBC? If not, then it should not add to hsbc

Some ad-block rulesets include liveperson.net, so I have to whitelist it to use the HSBC mobile app. But I think you're correct, it’s probably better to create a standalone LivePerson ruleset for this.

> 2. Why do you name it `schk` instead of `standardchartered`? It seems to be a UK company
> 
> 3. Please change `online.standardchartered.com` to `standardchartered.com`

I have completed the refactoring

MkQtS (approved):

PR #3215, opened 2026-01-24
Update twitch

hydraponique:
Adds new sub-domains of cloudfront.net which connected with Twitch

### DIFF
--- a/data/twitch
+++ b/data/twitch
@@ -6,6 +6,8 @@ twitch.tv
 twitchcdn.net
 twitchsvc.net
 
+full:d13j2sbsb4ncgj.cloudfront.net
+full:d186rixkn5ruba.cloudfront.net
 full:d1g1f25tn8m2e6.cloudfront.net
 full:d1m7jfoe9zdc1j.cloudfront.net
 full:d1mhjrowxxagfy.cloudfront.net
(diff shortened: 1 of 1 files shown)


EpLiar:
Thank you for your contribution, but can you provide these domains' source?

hydraponique:
> Thank you for your contribution, but can you provide these domains' source?

taked from Wireshark (PC) + checked by PCAPDroid (Android)

rootmelo92118:
> > Thank you for your contribution, but can you provide these domains' source?
> 
> taked from Wireshark (PC) + checked by PCAPDroid (Android)

Do you have any record of the file or a screenshot?

ghost:
Simply accept FQDNs which have proof & deny else.

---

Yes — at least **d186rixkn5ruba.cloudfront.net** is used by Twitch, because it appears as the CloudFront target behind Twitch CDN hostnames like `static-cdn.jtvnw.net` / `jtvnw.twitchcdn.net`. [github](https://github.com/DevilXD/TwitchDropsMiner/issues/386)

For the full list you gave, these are **very likely Twitch-related CDN hostnames** rather than general-purpose public Twitch domains: [netify](https://www.netify.ai/resources/applications/twitch)
- d13j2sbsb4ncgj.cloudfront.net
- d186rixkn5ruba.cloudfront.net
- d2mb634i1sflda.cloudfront.net
- d36mepituis1gg.cloudfront.net
- d3m1uefep57n8q.cloudfront.net
- dxtkri2c61io4.cloudfront.net

Why this is the likely answer: Twitch uses Amazon CloudFront heavily for delivery of assets and media, and Twitch CDN hostnames often resolve through random-looking CloudFront subdomains. Randomized CloudFront hostnames alone do not prove ownership, but the pattern is consistent with Twitch infrastructure. [netify](https://www.netify.ai/resources/hostnames/b.twitchcdn.net)

So the practical answer is: **yes, these are Twitch-associated domains in the CDN sense, though they are CloudFront hostnames and not the main Twitch site itself**. [github](https://github.com/DevilXD/TwitchDropsMiner/issues/386)

PR #3219, opened 2026-01-27
add Samsung OneUI update server

futuretech6:
I'm not sure if the `@update` tag is appropriate for this project.

Sometimes I want to allow this domain to update my phone, but other times I want to block it so I can stay on my current version. So a standalone tag can help separate the update server from other servers.

### DIFF
--- a/data/samsung
+++ b/data/samsung
@@ -8,6 +8,7 @@ galaxyappstore.com
 galaxymobile.jp
 game-platform.net
 knoxemm.com
+ospserver.net
 samsung.com
 samsungads.com @ads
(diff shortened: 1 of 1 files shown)


MkQtS:
> I'm not sure if the `@update` tag is appropriate for this project.
> 
> Sometimes I want to allow this domain to update my phone, but other times I want to block it so I can stay on my current version. So a standalone tag can help separate the update server from other servers.

No. It's not appropriate to add such a new attribute in public geosite. You should just change your own routing config for these specific domains to meet your needs.

PR #3262, opened 2026-02-10
kinopub: add kpdl.link

unixty:
Delete regexp, add short link for downloading apk

### DIFF
--- a/data/kinopub
+++ b/data/kinopub
@@ -2,6 +2,8 @@ kino.pub
 kinopub.online
 kpdl.link
 
+# Mirror sites
+gfw.ovh # sub domains mirror
 mos-gorsud.co # kinopub domain to generate a mirror site through gfw.ovh
 
 # kinopub CDN servers
@@ -9,4 +11,5 @@ cdn-service.space
 cdn2cdn.com
 cdn2site.com
 pushbr.com # poster images CDN
-cdntogo.net
+
+regexp:(\w+)-static-[0-9]+\.cdntogo\.net$


MkQtS:
Why did you delete `gfw.ovh`? And what's your data source for the new domains?

unixty:
> Why did you delete `gfw.ovh`? And what's your data source for the new domains?

personal mirrors doesn't need to be in domains list, it's personal to bypass blocks. Developer console on the site is data source

MkQtS:
@DanielLavrushin @mazixs Could you help to review this?

DanielLavrushin:
@MkQtS, sure, here it is:

1. `gfw.ovh` is a kinopub proxy domain that generates personal subdomains (e.g. abcdef.gfw.ovh). Removing this might result in not correctly proxying kinopub traffic for those who expect it to route via VPS. This is not about public/personal — this is about what is used by kino.pub services.

2. Removing the regex and keeping a broader `cdntogo.net` seems wrong. [cdntogo.net](https://cdntogo.net/) is a shared CDN provider and may serve unrelated third-party content. The regex ensures we only proxy kinopub-specific requests (the kinopub pattern is well known — e.g. `1a5af214-d010-4ccc-b666-eba61130c0db.ams-static-02.cdntogo.net`). I vote for keeping the regex as is.

3. `kpdl.link` is a valid kino.pub link, so yes, it makes sense to add it.

DanielLavrushin:
@MkQtS actually, regarding `gwf.ovh` - @unixty  might be right, In fact kino.pub randomly generates the 2d level domains as well, so no need to have only gwt.ovh in the list, I believe it  can be also removed.

So the only concern is about the `cdntogo.net`, but I can see the other people have already added a whide range of other cdns. So you decide in the end :).

MkQtS:
Since `gfw.ovh` is already there, we can keep it to avoid disrupting existing configurations. I prefer not to change `gfw.ovh` and `cdntogo.net`, just add the new domain `kpdl.link`.

@DanielLavrushin, thanks for your help. If you don't mind, please help to check other ru related PRs :)

DanielLavrushin:
@MkQtS Sure, glad to help! 
In the future, if you need any help with these (RU-related PRs), feel free to tag me like you did here — I don't often check PRs in this repo, but I'm happy to take a look when poked.

PR #3265, opened 2026-02-10
geolocation-!cn: include more sub lists

MkQtS:
They are part of Chinese entities but **exclusively** serving for non-cn area

As https://github.com/v2fly/domain-list-community/pull/3198 moved them out from `category-xxx-cn`/`geolocation-cn`, this adds them into `category-xxx-!cn`/`geolocation-!cn`

If there's no objection, I'll merge this in a week

### DIFF
--- a/data/category-non-domestic-cn
+++ b/data/category-non-domestic-cn
@@ -1,0 +1,31 @@
+# Part of Chinese entities but exclusively serving for non-cn area
+
+include:alibaba @!cn
+include:anker @!cn
+include:bilibili @!cn
+include:boc @!cn
+include:bytedance @!cn
+include:ccb @!cn
+include:chinamobile @!cn
+include:chinatelecom @!cn
+include:chinaunicom @!cn
+include:citic @!cn
+include:cmb @!cn
+include:ctrip @!cn
+include:deepin @!cn
+include:dewu @!cn
+include:didi @!cn
+include:eastmoney @!cn
+include:huawei @!cn
+include:icbc @!cn
+include:ipip @!cn
+include:iqiyi @!cn
+include:jd @!cn
+include:oppo @!cn
+include:pingan @!cn
+include:sina @!cn
+include:tencent @!cn
+include:vivo @!cn
+include:xd @!cn
+include:xiaohongshu @!cn
+include:xiaomi @!cn
--- a/data/cn
+++ b/data/cn
@@ -2,3 +2,4 @@
 
 include:tld-cn
 include:geolocation-cn
+#include:category-non-domestic-cn
--- a/data/geolocation-!cn
+++ b/data/geolocation-!cn
@@ -1,5 +1,8 @@
 # This list contains domains that don't have access point in China mainland. This is opposite to geolocation-cn.
 
+# Part of Chinese entities but exclusively serving for non-cn area
+include:category-non-domestic-cn
+
 # AI Chat
 include:category-ai-!cn
 


nekolsd:
What are we doing here bro, the `category-companies` list states that

> This list contains Tech & Cyber companies and organizations outside China mainland.

Adding domains of China mainland-based companies that serve non-cn area into this list doesn't make sense. These domains still belong to Chinese companies, not foreign ones — this contradicts the list's own description and makes the categorization misleading.

This probably goes back to how the project categorized things in the first place — some lists may have been poorly named from the start. But that's not a reason to make it worse by adding more mismatched entries.

don't do this bro

MkQtS:
> Adding domains of China mainland-based companies that serve non-cn area into this list doesn't make sense

Well, I thought these domains are used by their foreign subsidiary which may considered as non-cn companies. `geolocation-!cn` is a bit too generic, and these lists match the `Tech & Cyber companies` topic, so I added them to `category-companies`.

I can add them directly in `geolocation-!cn`, or create a new list like `category-non-domestic-cn-companies`(the name may be shorter...). That's not a problem.

Beyond this naming/categorization detail, actually, I am more concerned about whether this change is reasonable/necessary. I think removing `@!cn` domain rules out from `category-xxx-cn` is quite reasonable. But after this, I found the change appears to diminish the value of these removed domains(originally `cn@!cn` domains) as they were removed from big categories but not added to new ones. I had a feeling that since I was the one who drove them away, it should be my responsibility to find them a place to go. So it may be a choice to add them into `geolocation-!cn` families.

TBH, I don't like the idea to add all `@cn` domains to `geolocation-cn` and all `@!cn` to `geolocation-!cn`. It doesn't make sense to treat google@cn domains as part of cn domains. Similarly, I'm puzzled about the necessity of adding `cn@!cn` into `-!cn` lists.

Do you have any suggestions about that? Feel free to share your ideas.

MkQtS:
How about handling it this way? @nekolsd

Meo597:
Under no circumstances should Chinese entities be included in `geolocation-!cn`, as they very likely possess both domestic and international IPs.
Once you do this, it will break a large number of existing routing policies.

MkQtS:
> Once you do this, it will break a large number of existing routing policies.

Could you explain this? I don't quite understand.

In my understanding, we mark one domain with `@!cn` only if the domain resolves IPs outside cn, or the service is exclusively serving for non-cn area. If any domain belongs to a Chinese entity and resolves to cn IP/serves for cn area, then it should not be marked with `@!cn`.

Actually, some downstream projects, Loyalsoldier/MetacubeX/SagerNet, are even more aggressive. They already added `cn@!cn` to `geolocation-!cn` and `!cn@cn` to `geolocation-cn`. I didn't see much dissatisfaction from users.

Currently, this repo has moved `cn@!cn` out from `geolocation-cn` and `category-xxx-cn`. This PR is going to add these `cn@!cn` rules to `geolocation-!cn`. No one expressed a strong preference for this change, including myself, I have no intention to merge it now. I'd like to keep this PR open to receive more opinions.

Meo597:
Usually, people consider `geolocation-!cn` and `cn` to represent non-mainland and mainland entities, respectively.
Therefore, `loc-!cn` is often seen as being used by the GFW list, especially when users have little VPS traffic or poor router.

Once overseas businesses of Chinese entities are merged into `loc-!cn`, it might change user intent, as they did not originally want these domains to pass through the proxy.

MkQtS:
> `loc-!cn` is often seen as being used by the GFW list

`geolocation-!cn` is not GFW list. There are many cn available domains, like apple, microsoft, google@cn. It is unreasonable to expect `geolocation-!cn` to achieve the same effect as GFW list.

> it might change user intent, as they did not originally want these domains to pass through the proxy.

If anyone does not want to proxy all `!cn` traffic, then he should not proxy the whole `geolocation-!cn`.

I think we should do things right, not cater to existing improper perceptions/expectations.

Meo597:
This will trigger large-scale configuration changes downstream.

Because some people want Chinese entities to connect directly via Chinese IPs, even for overseas services, especially for things like online banking.
Now they have to relearn the new semantics and adjust the order of the rules.

Placing `loc-!cn@cn` into `cn` does not have this effect; we generally trust the moral standards of international entities.

ghost:
> Because some people want Chinese entities to connect directly via Chinese IPs, even for overseas services, especially for things like online banking.

So true.

_If it ain't broke, don't fix it_

xiaoxizi-lyx:
> > Once you do this, it will break a large number of existing routing policies.
> 
> Could you explain this? I don't quite understand.
> 
> In my understanding, we mark one domain with `@!cn` only if the domain resolves IPs outside cn, or the service is exclusively serving for non-cn area. If any domain belongs to a Chinese entity and resolves to cn IP/serves for cn area, then it should not be marked with `@!cn`.
> 
> Actually, some downstream projects, Loyalsoldier/MetacubeX/SagerNet, are even more aggressive. They already added `cn@!cn` to `geolocation-!cn` and `!cn@cn` to `geolocation-cn`. I didn't see much dissatisfaction from users.
> 
> Currently, this repo has moved `cn@!cn` out from `geolocation-cn` and `category-xxx-cn`. This PR is going to add these `cn@!cn` rules to `geolocation-!cn`. No one expressed a strong preference for this change, including myself, I have no intention to merge it now. I'd like to keep this PR open to receive more opinions.

I'd like to add a few points to this discussion:

**1. If the goal is truly to route based on where the server is actually located, that requires IP-level classification, not domain-level classification**

If what users actually want is "route based on the server's real geographic location," the most rigorous approach would be to resolve all domains to IPs first, then decide routing based on the IP's actual region — rather than giving a domain a fixed regional label. A static, domain-level classification is at best an approximation of "which region this entity tends to serve," not a precise reflection of "where this specific connection is actually going right now."

**2. On this statement:**

> In my understanding, we mark one domain with `@!cn` only if the domain resolves IPs outside cn, or the service is exclusively serving for non-cn area. If any domain belongs to a Chinese entity and resolves to cn IP/serves for cn area, then it should not be marked with `@!cn`.

If we follow this standard strictly, two practical problems arise:

- **Problem 1 — this breaks the user's original expectation of routing by entity as a whole, forcing an unwanted granular split.** Consider a user whose original intent for a given entity is: "I want to reach this entity entirely through my real IP" — based on trust in, or a judgment about, the entity as a whole, rather than evaluating each subdomain individually. But under MkQtS's approach, once `us.example.com` (which this entity uses exclusively to serve US users) gets classified into `!cn`, the user can no longer achieve that "route by entity as a whole" intent. Instead, they're forced into a finer split they may never have wanted: "use my real IP when accessing this entity's China-facing domains, but go through the proxy when accessing its domains serving other regions." This kind of granularity may not be what the user actually wants — especially since the current entity-based classification convention has been in place for a long time, and a large number of users and downstream projects have built their rules around this exact mental model. Suddenly changing the classification logic risks silently proxying a large volume of domains that users never intended to be proxied.

- **Problem 2 — some entities could use paired domains (e.g. `cn.example.com` and `us.example.com`) to correlate a user's real IP with their proxy IP, effectively detecting proxy usage.** For example: a user wants to use their real IP for a given entity. But when the app/website loads, the client may simultaneously request both `cn.example.com` and `us.example.com`. Even though `us.example.com` genuinely resolves only to US IPs and exclusively serves the US region, once it's added to the `!cn` list and routed through the proxy — while `cn.example.com` goes through the user's real IP — the entity can observe both requests and correlate the user's real IP with their proxy IP, revealing that the user is behind a proxy. This runs counter to the user's intent and could expose them to risk. The reverse case holds too: if a user wants to *always* proxy traffic to a given entity, the entity can still fingerprint the user's real IP via requests to the `cn` domain.

**So my view is:** using "the region an IP resolves to" to help users decide whether a domain should be proxied is not something `domain-list-community` should be doing — it's the job of `geoip`. Resolve domains to IPs, then match against geoip. For a compliant Chinese company's `example.com`, none of its subdomains would realistically be blocked by the GFW, which makes it trivially cheap for such an entity to fingerprint a user's proxy IP via `us.example.com`.

That said, this approach isn't without trade-offs. It can't satisfy two legitimate user needs:

1. Users who genuinely want to use different IPs when accessing `us.example.com` versus `cn.example.com`;
2. Cases where the GFW actually blocks `us.example.com` but allows `cn.example.com` — in which case the user would run into connectivity issues accessing `us.example.com`.

Even so, I believe we should lean toward the more conservative approach, for three reasons:

1. It avoids exposing a user's "other IP" without their knowledge;
2. This project's domain classification isn't inherently tied to whether the GFW blocks a given domain, and the two shouldn't be conflated;
3. Advanced users with specific needs can always add their own custom rules and knowingly accept the associated exposure risk — that should be their own choice, not a default the project makes on behalf of all users.

MkQtS:
> A static, domain-level classification is at best an approximation of "which region this entity tends to serve," not a precise reflection of "where this specific connection is actually going right now."

Yes. And I don't think this PR breaks that.

>> In my understanding, we mark one domain with @!cn only if the domain resolves IPs outside cn, or the service is exclusively serving for non-cn area. If any domain belongs to a Chinese entity and resolves to cn IP/serves for cn area, then it should not be marked with @!cn.
> 
> If we follow this standard strictly, two practical problems arise:
>
> - **Problem 1 — this breaks the user's original expectation of routing by entity as a whole, forcing an unwanted granular split.** Consider a user whose original intent for a given entity is: "I want to reach this entity entirely through my real IP" — based on trust in, or a judgment about, the entity as a whole, rather than evaluating each subdomain individually.

The "original expectation" is not clear. Different users may have different expectations. Entities are different from categories. A list of a certain category/geolocation does not stand for one or multiple entities. In this PR, lists of entities are not changed at all. Only `geolocation-!cn` is affected. The question should be "Should we add non-cn domains of China entities into lists of `!cn` categories?"

> But under MkQtS's approach, once us.example.com (which this entity uses exclusively to serve US users) gets classified into !cn, the user can no longer achieve that "route by entity as a whole" intent.

Users can still achieve the "route by entity as a whole" intent, just add an explicit routing rule in their configuration before the `geo-!cn` rule. This is always recommended to achieve any intents.

> Instead, they're forced into a finer split they may never have wanted: "use my real IP when accessing this entity's China-facing domains, but go through the proxy when accessing its domains serving other regions."

The new *finer split* may be the **exact intents of some other users**. Both intents are not wrong. Nobody is forced to route any traffic against their intents. If this PR violates some users' intents, then I would say the original situation violates other users' intents. The reason does not make sense. Once again, the question should be "Should we add non-cn domains of China entities into lists of `!cn` categories?"

A classic example is tiktok and bytedance. If you visit tiktok directly in China, then it's not available at all. Another example is foreign games bought by tencent, the game may still be unavailable after the legal ownership changes. If `geolocation-cn` stands for collections of Chinese entities and `geolocation-!cn` must exclude them, then tiktok and the foreign games of tencent should be included in `geolocation-cn`, not `geolocation-!cn`. Is this the behavior most users expect?

> This kind of granularity may not be what the user actually wants — especially since the current entity-based classification convention has been in place for a long time, and a large number of users and downstream projects have built their rules around this exact mental model.

The *entity-based classification* in category lists is somehow caused by the limit of old `main.go`, the old `include` syntax old supports complete inclusions, no filters for attrs. So the lists of categories were somehow *limited* by the included lists of entities. This is why tiktok was included in both `geolocation-cn` and `geolocation-!cn` before, see #604, which was a long-existing problem that should be fixed. After the refactoring of `main.go`, selective inclusion is available now. So we can keep the ownership of tiktok and bytedance (ignore the tiktok seal deal here) while keep tiktok out of `geolocation-cn`.

For downstream projects:

> Actually, some downstream projects, Loyalsoldier/MetacubeX/SagerNet, are even more aggressive. They already added `cn@!cn` to `geolocation-!cn` and `!cn@cn` to `geolocation-cn`. I didn't see much dissatisfaction from users.

The literal sober meaning of `geolocation-!cn` should be domains that located outside China, and this PR matches the meaning. And many users also write their configurations based on the literal sober meaning. I think this is the goal we should pursue.

> Suddenly changing the classification logic risks silently proxying a large volume of domains that users never intended to be proxied.

Yes. This would change some default behavior. So we are disscussing here.

> - **Problem 2 — some entities could use paired domains (e.g. `cn.example.com` and `us.example.com`) to correlate a user's real IP with their proxy IP, effectively detecting proxy usage.** For example: a user wants to use their real IP for a given entity. But when the app/website loads, the client may simultaneously request both `cn.example.com` and `us.example.com`. Even though `us.example.com` genuinely resolves only to US IPs and exclusively serves the US region, once it's added to the !cn list and routed through the proxy — while `cn.example.com` goes through the user's real IP — the entity can observe both requests and correlate the user's real IP with their proxy IP, revealing that the user is behind a proxy. This runs counter to the user's intent and could expose them to risk. The reverse case holds too: if a user wants to always proxy traffic to a given entity, the entity can still fingerprint the user's real IP via requests to the `cn` domain.

Services with tricky proxy detection should be handled with additional attention. Since you don't want `example.com` to get your real IP in China, then it should be a foreign service mainly listed in `geolocation-!cn`. `cn.example.com` may get marked with `@cn` attr (we can remove `@cn` attr based on user's reasonable requests), and we have no need to mark `us.example.com` with `@!cn` attr. It's the `!cn@cn` issue, like "Should we include `google@cn` domains into cn lists", not the `cn@!cn` issue here. BTW, I'm not intersted in adding `!cn@cn` domains into `geolocation-cn`.

For `cn@!cn` domains, the risk should be the leakage of the proxy IP. Unless you just proxy limited known domains, and leave other domains direct, the leakage of proxy IP is hard to avoid.

- For unlisted domains, many users default to proxy them.
- There're 10k+ existing domains in `geolocation-!cn`, Chinese entities can buy a certain domain easily and deploy the detection and we can hardly found.

So I don't this is a problem we can avoid.

PR #3267, opened 2026-02-11
xiaohongshu: add more rednote domains

EpLiar:


EpLiar:
source:
image

MkQtS:
These rednote domains do not resolve to cn IPs. So they should be marked as `@!cn`? (I found `rednotecdn.com` in my logs, it's ICP filed. I didn't check the corresponding IP...)

EpLiar:
> These rednote domains do not resolve to cn IPs. So they should be marked as `@!cn`? 

Thank you for the reminder. I checked these domains at https://dnschecker.org/ and confirmed that none of their resolved addresses are Chinese IPs.

PR #3275, opened 2026-02-14
Add domains lists for eufy, Petlibro, ZScaler

echernyavskiy:


MkQtS:
1. Subdomains are unnecessary as they are overridden by the parent domain. e.g. `[domain:]eufy.com` overrides `[domain:]app-news-eu-pr.eufy.com`
2. Adjacent rules should be sorted alphabetically
3. Newly added list should be included in relevant categories if possible

MkQtS:
For eufy, use `geosite:anker`; zscaler is added; petlibro is not added as there're too many third-party domains

PR #3313, opened 2026-02-28
Porn updates

ghost:
- Create "illgames" file
  - FYI, "illusion" is already gone. Why not remove it?
- Update DLsite
- Create "doujin" file
  - To list small porngame websites. There are many more so making file for each website is not ideal.
- Create "category-urlshortner"

### DIFF
--- a/data/category-games-!cn
+++ b/data/category-games-!cn
@@ -13,7 +13,7 @@ include:gaijin
 include:garena
 include:gog
 include:hoyoverse
-include:illusion
+# Dead include:illusion
 include:itchio
 include:lowiro
 include:modrinth
--- a/data/category-porn
+++ b/data/category-porn
@@ -25,7 +25,7 @@ include:hentaichen
 include:hentaivn
 include:heyzo
 include:hooligapps
-#include:illusion
+# Dead include:illusion
 include:illusion-nonofficial
 include:illgames
 include:jable
--- a/data/geolocation-!cn
+++ b/data/geolocation-!cn
@@ -156,6 +156,7 @@ include:category-enhance-gaming
 include:category-mobile-repair
 include:category-ntp
 include:category-password-management
+include:category-urlshortner
 include:category-web-archive
 include:decryptipastore # iOS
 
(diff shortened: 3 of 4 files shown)


MkQtS:
- `doujin` is not a good name. It's not reasonable to combine these irrelevant domains into a weird list.
- Why do we need the new `category-urlshortner`?

ghost:
> * It's not reasonable to combine these irrelevant domains into a weird list.

This repository already have a `porn website` list. I added _some_ `doujin` websites to the list.

Take a look here for game examples: (NSFW, obviously)
- https://www.dlsite.com/maniax/?locale=zh_CN
- https://news.murax2.com/2024/01/05/dlsite-featured-h-games/

> `doujin` is not a good name. 

IDK what you call "doujin" in Chinese, but it become an English word: https://en.wikipedia.org/wiki/Doujin
How about `porn-games` then?

>     * Why do we need the new `category-urlshortner`?

2 reasons:
- You already have some URLshortners.
- There are at least 3 shortener lists on github, mainly for blocking purpose. By having meta category, `category-urlshortner`, user will have a choice to decide whether to proxy/drop shortners for privacy;
  - `URL shortener blocking provides an important security layer by preventing access to potentially dangerous shortened links`
  - e.g., URL shortener list: https://github.com/hagezi/dns-blocklists?tab=readme-ov-file#urlshortener

MkQtS:
1. It' not about Chinese. If I understand the word correctly, `doujin` means a group of fellow enthusiast, and it seems not directly related to porn or games. As for the usage here, it stands for a group/collection/category and not a company/organization/entity, so it would be better to be named as `category-xxx`. `category-porn-games` would be a qualified list name, but also related to next issue.
2. This repo is mainly for proxy usage. We have plenty of categories and some of them appear somewhat arbitrary, lacking a clear standard for whether a category is justified or which categories should be created.

Adding new entities and adding domains to existing entities/categories are fine. But I think we should avoid creating too many new useless categories.

ghost:
Ok I simply removed the file. Other commits are ready to merge.

PR #3319, opened 2026-03-02
Add plex domain list for routing

rz467fzs7d:
Add data/plex with Plex domains. Parent domains cover all subdomains per project guidelines. v4.plex.tv and domain:plex.direct are marked with @direct attribute as they are required for Plex to work properly.

Plex is a popular media server service and deserves its own geosite category. 

Usage:
- geosite:plex - matches all Plex domains
- geosite:plex@direct - matches only direct connection domains

### DIFF
--- a/data/category-companies
+++ b/data/category-companies
@@ -66,6 +66,7 @@ include:openweather
 include:oracle
 include:panasonic
 include:pccw
+include:plex
 include:qnap
 include:qualcomm
 include:qwant
--- a/data/plex
+++ b/data/plex
@@ -1,0 +1,14 @@
+# Plex Media Server domains
+# Basic domains for Plex services
+plex.com
+plex.services
+plex.tv
+plexapp.com
+
+# Note for Plex users
+# * Plex's mechanism requires certain domains to go through a direct connection.
+# * Using a proxy for these domains may cause the system to malfunction.
+# * If you wish to access plex domains through a proxy, remember to ensure the following domains remain directly connected.
+# * Related domains:
+# plex.direct
+# v4.plex.tv


MkQtS:
The new attr `direct` is not reasonable. `plex.direct` is already added in `data/private`. What makes `v4.plex.tv` special?

BTW, @Aethersailor, are you sure that `plex.direct` and `full:plex.llon.net` should be added in `category-emby`?

https://github.com/v2fly/domain-list-community/blob/6b16be40206f31857dd197a92c03f51df2734a90/data/category-emby#L27-L29

rz467fzs7d:
The reason I added @direct attribute is to support use cases where certain Plex domains need direct connection:                                                       
                                                                                                                                                                        
  - Plex server performs online status detection by checking v4.plex.tv                                                                                                 
  - If this domain goes through proxy, the server's public IP will be incorrectly identified as the proxy exit IP, causing Plex to report "cannot connect directly"

  Using geosite:plex@direct allows users to route only the essential domains through direct rules while keeping other Plex domains through proxy.

  However, I noticed that the current implementation doesn't include @direct. Is there a better approach to achieve this? Would it make more sense to use a different
  attribute name or a separate category?

> The new attr `direct` is not reasonable. `plex.direct` is already added in `data/private`. What makes `v4.plex.tv` special?
> 
> BTW, @Aethersailor, are you sure that `plex.direct` and `full:plex.llon.net` should be added in `category-emby`?
> 
> https://github.com/v2fly/domain-list-community/blob/6b16be40206f31857dd197a92c03f51df2734a90/data/category-emby#L27-L29

MkQtS:
`geosite:private` contains `[domain:]plex.direct` as the `*.plex.direct` resolves to private IPs. The case is different for `v4.plex.tv` as it resolves to a public IP. So we should not add `v4.plex.tv` to `data/private`. (You can still add `plex.direct` to `data/plex` if necessary)

`[domain:]plex.tv` overrides `full:v4.plex.tv`. If they are in a same list and the later one doesn't contain any attr, it is actually meaningless and will not remain in the generated `dlc.dat`

Some cn companies, like bytedance, many have some domains that not serving for cn, like tiktok, then we add `@!cn` attr to them. But the case here is different.

You may create two file: `data/plex` and `data/plex-direct`(include the later one in the former one if neceessary) to handle such case.

But, the question is: must `v4.plex.tv` be proxied and others be directed? Or is there any explicit differences between `plex-proxy` and `plex-direct`? If someone set his/her remote server for proxy also as the plex storage server, does he/she need to proxy both `plex-proxy` and `plex-direct`?

If there's no explicit differences between `plex-proxy` and `plex-direct`, then just use one `data/plex`. You can add comments for plex users to advise them to properly handle `v4.plex.tv` traffics.

rz467fzs7d:
You're right. I've split it into two files:                                                                                                                           
   
  - `data/plex`: basic Plex domains (can be proxied for better experience)                                                                                              
  - `data/plex-direct`: v4.plex.tv must use direct connection                                                                                                         
                                               
  The reason: v4.plex.tv is used for Plex server's online status detection. If this domain goes through proxy, the server's public IP will be incorrectly identified as the proxy exit IP, causing Plex to report "cannot connect directly".

  I apologize for the confusion in my previous explanation. Let me clarify: **v4.plex.tv must use direct connection** - this is required for Plex server to work  properly. Other Plex domains can be proxied if needed.

  This allows users to:
  - Use `geosite:plex` for basic Plex domains (proxy recommended)
  - Use `geosite:plex-direct` for direct-only domains (required for v4.plex.tv)

  Does this approach make sense?

rz467fzs7d:
> `geosite:private` contains `[domain:]plex.direct` as the `*.plex.direct` resolves to private IPs. The case is different for `v4.plex.tv` as it resolves to a public IP. So we should not add `v4.plex.tv` to `data/private`. (You can still add `plex.direct` to `data/plex` if necessary)
> 
> `[domain:]plex.tv` overrides `full:v4.plex.tv`. If they are in a same list and the later one doesn't contain any attr, it is actually meaningless and will not remain in the generated `dlc.dat`
> 
> Some cn companies, like bytedance, many have some domains that not serving for cn, like tiktok, then we add `@!cn` attr to them. But the case here is different.
> 
> You may create two file: `data/plex` and `data/plex-direct`(include the later one in the former one if neceessary) to handle such case.
> 
> But, the question is: must `v4.plex.tv` be proxied and others be directed? Or is there any explicit differences between `plex-proxy` and `plex-direct`? If someone set his/her remote server for proxy also as the plex storage server, does he/she need to proxy both `plex-proxy` and `plex-direct`?
> 
> If there's no explicit differences between `plex-proxy` and `plex-direct`, then just use one `data/plex`. You can add comments for plex users to advise them to properly handle `v4.plex.tv` traffics.

My original idea was to follow the same syntax as apple@cn: mark domains that need direct connection with @direct attribute in a single data/plex file. 

This way  all Plex domains can be maintained in one file, which is more intuitive for maintenance. When used, geosite:plex-direct goes first for direct connection, then  geosite:plex for proxy. 
However, this would introduce a new tag @direct that hasn't been used before. 

If you also think @direct is a better approach, I'd be  happy to update the PR accordingly.

MkQtS:
Plex is partially blocked by GFW, but to make plex works, some plex domain (`v4.plex.tv`) cannot be proxied. So you want to keep certain plex traffic direct but proxy most other domains. It's a conflict between GFW and plex's mechanism. As for a free network, there's no need to distinguish plex-proxy and plex-direct traffic. Is this correct?

To address the problem here in geosite, the above solutions, no matter creating new attr `@direct` or adding two lists (`plex` and `plex-direct`), are somewhat strange or distorted. I'm not sure whether we should do so.

rz467fzs7d:
> Plex is partially blocked by GFW, but to make plex works, some plex domain (`v4.plex.tv`) cannot be proxied. So you want to keep certain plex traffic direct but proxy most other domains. It's a conflict between GFW and plex's mechanism. As for a free network, there's no need to distinguish plex-proxy and plex-direct traffic. Is this correct?
> 
> To address the problem here in geosite, the above solutions, no matter creating new attr `@direct` or adding two lists (`plex` and `plex-direct`), are somewhat strange or distorted. I'm not sure whether we should do so.

My latest solution is to use @cn attribute to mark v4.plex.tv. Actually, I noticed that other -cn categories don't strictly mean pure Chinese domains, although some might be; but these domains are definitely accessible in China. Usually in geosite declarations, using geosite:xxx-cn means the domain can be directly  connected.                                                                                                                                                      
   
  So my final solution is: add @cn after v4.plex.tv to ensure all Plex domains and this special one can be maintained in one file.                                
                                                                                                                                                                
  Plex Media Server domains                    

  plex.com
  plex.services
  plex.tv
  plexapp.com
  v4.plex.tv @cn

  Usage:
  - `geosite:plex` - all Plex domains
  - `geosite:plex@cn` - v4.plex.tv for direct connection

  I appreciate your careful review and welcome any direct modifications to my code if needed.

MkQtS:
I'm not sure... What do you think about it? @rootmelo92118

rz467fzs7d:
> I'm not sure... What do you think about it? @rootmelo92118

This follows the same pattern used elsewhere in the project. 
For example, in [data/microsoft-pki](url), www.microsoft.com is marked with @cn even though it's not a .cn domain — it's an international domain that happens to be accessible from China (via CDN or direct connection).    
                                                                         
Similarly, v4.plex.tv is an international domain that is accessible from China, so @cn is appropriate here.

MkQtS:
See the README:

> Domains and services that originate from outside China mainland but have access point in China mainland, may be marked by attribute `@cn`

`www.microsoft.com` resolves to China IPs, so it can be marked with `@cn`. `v4.plex.tv` resolves to amazon in US.

Adding `@cn` attr to `v4.plex.tv` is somewhat a solution for the problem here, but also may be an abuse of attributes.

rz467fzs7d:
> See the README:
> 
> > Domains and services that originate from outside China mainland but have access point in China mainland, may be marked by attribute `@cn`
> 
> `www.microsoft.com` resolves to China IPs, so it can be marked with `@cn`. `v4.plex.tv` resolves to amazon in US.
> 
> Adding `@cn` attr to `v4.plex.tv` is somewhat a solution for the problem here, but also may be an abuse of attributes.

Now it's a pure collection of plex domain names

MkQtS:
I've added some notes and included plex in `category-companies`. Please check it. If there's no problem, I'll merge it. Feel free to comment or open another PR if you find a better solution.

rz467fzs7d:
> I've added some notes and included plex in `category-companies`. Please check it. If there's no problem, I'll merge it. Feel free to comment or open another PR if you find a better solution.

Looks good to me, thanks for the patient review!

PR #3350, opened 2026-03-11
anthropic: add clau.de

direct808:
Сli tool accesses the domain clau.de


EpLiar (changes requested):
clau.de is ok for this file, but you should sort it

PR #3367, opened 2026-03-18
main.go: support to generate multiple custom dats

MkQtS:
As the project data continues to accumulate, the `dlc.dat` is growing larger. Although some lists are rarely used in practice, and some exist only as aliases for compatibility, we will probably always retain them. In the generated `dlc.dat`, because nested inclusions between lists are resolved and expanded, a single rule may actually be stored three or four times, taking up additional space.

On the other hand, unless the source data is modified or additional custom code is written, this project does not support to generate custom dat files. While this ensures consistency in the final generated dat file, it also limits advanced users to customize the output.

This PR allows users to decide for themselves, through custom configuration files, which lists to retain or remove from the final database. (If there is broad consensus within the community, we might release an official "lite" version.) This approach does not require modifying the source data (though users are free to do so) and carries no risk of breaking the inclusion chain. As long as the source data is consistent, any list remains in the trimed dat contains the same rules comparing to the list in full dat.


MkQtS:
If this get merged, users can create custom .dat files easily.

Here's an example:

```shell
#!/bin/bash

#cd domain-list-community || exit
mkdir -p ./data/playground

# create new list: diy-block
cat > ./data/playground/diy-block <<- EOF
	include:category-ads-all
	include:category-httpdns-cn
	any-domain-i-want-to-block.com
EOF

# create new list: diy-direct
cat > ./data/playground/diy-direct <<- EOF
	include:apple
	include:microsoft
	include:geolocation-cn
	include:geolocation-!cn @cn
	include:private
	any-domain-i-want-to-direct.com
EOF

# create new list: diy-proxy
cat > ./data/playground/diy-proxy <<- EOF
	include:geolocation-!cn @-cn
	any-domain-i-want-to-proxy.com
EOF

# to be mixed with data/private
cat > ./data/playground/private <<- EOF
	internal-domain-of-my-home
	internal-domain-of-my-company
EOF

cat > ./dat_diy.json << EOF
[
	{
		"name": "custom_full.dat",
		"mode": "all"
	},
	{
		"name": "custom_tiny.dat",
		"mode": "allowlist",
		"lists": [
			"diy-block",
			"diy-direct",
			"diy-proxy",
			"category-ai-!cn"
		]
	},
	{
		"name": "custom_nodiy.dat",
		"mode": "denylist",
		"lists": [
			"diy-block",
			"diy-direct",
			"diy-proxy"
		]
	}
]
EOF

go run ./main.go --datapath='./data' --datprofile='./dat_diy.json'

```

@database64128, thanks for your review :) Feel free to share any ideas.

Loyalsoldier:
It's great to support to generate customized dat files for advanced users, but I don't like the idea of generating `dlc_nobackward.dat` by default. IMO, the `dlc.dat` is small and good to go.

MkQtS:
> I don't like the idea of generating dlc_nobackward.dat by default

`dlc_nobackward.dat` or the file `datprofile.json` is now only an example. I'm not going to release `dlc_nobackward.dat`. Maybe I should remove this file.

PR #3373, opened 2026-03-19
category-dev: add ziglang.org

infirms:


MkQtS:
Please do not create new list for a single domain. Just add `ziglang.org` in `category-dev`.

infirms:
fixed

PR #3414, opened 2026-04-07
VK Group services update

yetanotherqubick:
As described in the title I decided to merge all VK Group services into a single file. VK Group has a shitload of domains, that are sometimes used interchangeably across its entire ecosystem. It's mostly related to CDNs, but almost all of their services also communicate with `vk.com` in some way. Thus I think merging would be justified.

It *might* be possible to categorize them all, but I doubt it's necessary at this point.

Besides merging I've added a lot of domains taken mainly from `vk.company/en/company/about/`, `vk.company/en/projects/`, `vk.company/en/company/business/`, and a bit from Wikipedia. I also used whois to confirm that some of the domains are indeed related to VK Group.

Sources of advertising/tracking domains (many thanks to them all yet again!):
https://github.com/uBlockOrigin/uAssets
https://easylist.to/
https://pgl.yoyo.org/adservers/
https://forums.lanik.us/viewforum.php?f=102
Some of them were added simply by lurking for subdomains.

I think it *shouldn't* cause any problems.


MkQtS:
Removing these lists may break some existing configurations. Why do we need to remove them?

yetanotherqubick:
> Removing these lists may break some existing configurations. Why do we need to remove them?

I thought about this. We could:

1. Include `mailru-group` in these files for compatibility.
2. Create a separate `mailru-group-cdn`, add deleted files back and include these CDNs in every appropriate file.

I think the latter option might be better.

yetanotherqubick:
@MkQtS what do you think?

MkQtS:
If this get merged, dzen, mailru, ok and vk all become alias list of mailru-group. This is somehow inconsistent with the name of the list, user's expectations or current practice.

We can not clearly categorize domains of a large company into different lists, because some internal infra domains are widely used among all their business. There are some similar cases, like `tiktok` uses domains in `bytedance`, `google-deepmind`(Gemini) uses domains in `google`. Current practice is adding those exclusively used domains into these small lists, and generic domains directly into the parent list of the entire company. While this is not perfect, I don't know any better solution.

I will probably not merge this unless more users thought this is necessary.

yetanotherqubick:
@MkQtS are attributes preserved when I include one file into another? I mean, should I also include `dzen`, `ok` and `vk` info `category-ads`?

Also while I'm fine with `dzen`, `ok` and `vk` as a separate files (perhaps they really should stay that way), I still think `mailru` could become alias for `mailru-group`.

MkQtS:
Attributes are part of the domain rule, they are always preserved.

> should I also include `dzen`, `ok` and `vk` info `category-ads`?

No need to do so. Rules in these child lists were included into `mailru-group`, so `include:mailru-group @ads` contains all rules with `@ads` attr from child lists. You may try `go run .\main.go --exportlists='mailru-group,category-ads'` to check it.

yetanotherqubick:
I hope no changes would be needed now :)

MkQtS:
I'll review it tomorrow. Good night:)

ghost:
> If this get merged, dzen, mailru, ok and vk all become alias list of mailru-group

Nice - so if I want to block VK and its parent company I simply block `mailru-group`

yetanotherqubick:
> > If this get merged, dzen, mailru, ok and vk all become alias list of mailru-group
> 
> Nice - so if I want to block VK and its parent company I simply block `mailru-group`

You can do this now too: `mailru-group` includes `dzen`, `ok` and `vk` already. I was willing to merge these, because there's almost no sense and structure in how VK (Group) creates and utilizes its domains. I'd say it's purely chaotic.

MkQtS (approved):

Loyalsoldier:
Please remove prefix `domain:`

yetanotherqubick:
> Please remove prefix `domain:`

Sure, but may I ask why?

ghost:
> > Please remove prefix `domain:`
> 
> Sure, but may I ask why?

IIUC `*.basedomain` == `domain:basedomain`

PR #3424, opened 2026-04-09
Create category-ru-whitelisted

sngvy:
Due to restrictions imposed by Russian mobile operators and technology giants, this category is necessary to route traffic directly.


MkQtS:
1. There's already `category-ru`. What's the difference? Why is this new list necessary?
2. See the README: Purpose of this project

   > This project is not opinionated. In other words, it does NOT endorse, claim or imply that a domain should be blocked or proxied. It can be used to generate routing rules on demand.

   Some users may want google to be proxied while some users not. So users should write their own config. This repo is better to provide only lists with objective criterion
3. Use `include` to make the future maintaining eaisier (After there's enough consensus that this new list is reasonable)

sngvy:
1. The purpose of this category is to define the set of resources that require **strict direct routing**. It is known that, as of April 2026, Russian resources have begun deploying honeypots to detect proxies. Please consult independent sources (https://novayagazeta.eu/articles/2026/04/07/zadacha-naiti-vpn, https://meduza.io/news/2026/04/07/izvestiya-soobschili-chto-rossiyskie-marketpleysy-ogranichivayut-dostup-polzovatelyam-s-vklyuchennym-vpn). End users must therefore exclude this entire set of resources (_especially when whitelists are enabled on LTE networks_).
2. Proxying traffic to these resources exposes the proxy to the risk of detection and subsequent blocking. This brings us back to point 1.
3. I understand this.

yetanotherqubick (changes requested):
I don't think it's necessary to create `category-ru-whitelisted` since usually you would like to route directly *all* traffic inside Russia with `geosite:category-ru` and `geoip:ru`. Thus you decrease chance of detection even further (and use less proxy-server data/bandwidth).

I'd suggest to focus on including domains from non-`ru`/`su`/`рф` TLD zones that are used by these services, especially by these close to the state. I think however we should add whitelisted `ru`/`su`/`рф` domains explicitly to `category-ru`. Of course if these aren't present already elsewhere.

**UPD:** **Please**, categorize these at least with comments. Currently, it is difficult to navigate (and thus maintain) the file.

Resetand:
Я правильно понимаю что этот список не включает список всех аккредитованных IT компаний в России? Только те что из белых список

Контекст такой: Минцифры обязал все аккредитованные компании как минимум детектить и запрещать вход на их ресурсы, при это вполне возможно в будущем будет передача IP адресов VPN-ов РНК для блокировки, чего хочется избегать, например обрубив доступ на эти ресурсы при включенном VPN

Насколько этот список подходит под эти цели? Стоит ли создать category-ru-it-accredited?

----

Am I correct in understanding that this list does not include all accredited IT companies in Russia? Only those from a whitelist?

The context is: the Ministry of Digital Development has required all accredited companies to at least detect and block access to their resources, and it’s quite possible that in the future VPN IP addresses will be shared with Roskomnadzor for blocking, which is something to avoid - for example, by cutting off access to these resources when a VPN is enabled.

How suitable is this list for these purposes? Is it worth creating a category-ru-it-accredited?

ghost:
Why make "whitelist"? What makes you think those sites are the good guys? Any websites can turn evil any time. It doesn't make much sense to make such list - what about "category-cn-whitelisted" then?

yetanotherqubick:
> Why make "whitelist"? What makes you think those sites are the good guys? Any websites can turn evil any time. It doesn't make much sense to make such list - what about "category-cn-whitelisted" then?

These aren't 'good'. These are whitelisted in Russia by the state. In case of the Internet shutdown these resources should be still available. It's mostly arbitrary, but to be whitelisted the Russian government demands from the companies to use only Russian infrastructure. Recently the government demanded from them to detect and block proxy, VPNs and other censorship circumvention attempts.

ACherepkov1989:
This list of domains will be useful in the current reality.
Have you checked these domains against whitelists? If not, you need to check them before adding them...
<details>
<summary>Spoiler</summary>
I found a list of some domains that should work, but they also need to be checked before adding...
[jinndi/geosite-cheburnet/geosite-cheburnet](https://github.com/jinndi/geosite-cheburnet/tree/main/geosite-cheburnet)
[hxehex/russia-mobile-internet-whitelist/whitelist.txt](https://github.com/hxehex/russia-mobile-internet-whitelist/blob/main/whitelist.txt)
</details>

sngvy:
Yes, they all work (test on MegaFon)

PR #3431, opened 2026-04-09
Update x5

yetanotherqubick:
New domain from `ru.pyaterochka.app.browser` network analysis, added ads section


MkQtS:
Why not adding `@ads` attr and include it in `category-ads`?

yetanotherqubick:
> Why not adding `@ads` attr and include it in `category-ads`?

Forgot to do this.

MkQtS (approved):

PR #3454, opened 2026-04-11
Update sber

yetanotherqubick:
Added two online payment and one tracking domains

### DIFF
--- a/data/sber
+++ b/data/sber
@@ -14,7 +14,7 @@ sbermarketing.ru @ads
 domain:ads.sber.ru @ads
 domain:clickstream.sberbank.ru @ads
 domain:error-tracking.megamarket.tech @ads
-domain:metrics.21-school.ru
+domain:metrics.21-school.ru @ads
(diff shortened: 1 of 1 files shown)


MkQtS:
Please do not use regexp for such simple cases. The regexp performance is not good. (And you should use use `^` and `$` to avoid match wrong or even malicious domains.) Similar for #3456

yetanotherqubick:
> Please do not use regexp for such simple cases. The regexp performance is not good. (And you should use use `^` and `$` to avoid match wrong or even malicious domains.) Similar for #3456

Sure, I'll keep that in mind. Should I revert changes for similar simple cases in other PRs? And which ones should be considered this 'simple'?

MkQtS:
> And which ones should be considered this 'simple'?

When plain domain/full is enough, avoid using keyword/regexp.

> Should I revert changes for similar simple cases in other PRs?

Yes. Just update this and #3456 for now.

MkQtS (approved):

PR #3456, opened 2026-04-11
category-bank-ru: added domains

yetanotherqubick:
Added Alfabank, BCS and MKB domains

### DIFF
--- a/data/category-bank-ru
+++ b/data/category-bank-ru
@@ -2,14 +2,21 @@ include:sber
 include:tbank-ru
 
 # Alfa Bank
-alfabank.ru
-alfabank.st
+alfadirect.ru
+myapelsin.ru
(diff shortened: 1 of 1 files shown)


data/category-bank-ru
```diff
@@ -4,13 +4,24 @@ include:tbank-ru
 # Alfa Bank
 alfabank.ru
 alfabank.st
+alfadirect.ru
+myapelsin.ru
+domain:alfabank.oavdo.amc
```
MkQtS, line 9: Just curious, why do you specify the type to `domain`? It's default type and can be omitted.

yetanotherqubick: Perhaps I thought that would bring more clarity. I also thought `domain` should be omitted primarily for second level domains.

MkQtS (approved):

PR #3469, opened 2026-04-14
Update Russian websites

ACherepkov1989:
category-gov-ru
- sev.ru: Remove a domain that doesn't belong to a federal city. The domain belongs to the energy company Sapsan Energia.

category-travel-ru
- Add Othello
- Add Fasten

category-bank-ru
- Add a comment for SberBank and T-Bank

mailru-group
- Add a missing mail.ru email domain
- Update comments for myMail and My.Games

### DIFF
--- a/data/category-bank-ru
+++ b/data/category-bank-ru
@@ -14,6 +14,7 @@ broker.ru
 gazprombank.com
 gazprombank.investments
 gazprombank.ru
+gid.ru
 gpb.ru
 
 # MKB
--- a/data/category-bank-ru
+++ b/data/category-bank-ru
@@ -1,6 +1,3 @@
-include:sber
-include:tbank-ru
-
 # Alfa Bank
 alfabank.ru
 alfabank.st
@@ -37,6 +34,12 @@ abr.ru
 
 # Russian Agricultural Bank
 rshb.ru
+
+# SberBank
+include:sber
+
+# T-Bank (Tinkoff) and RosBank
+include:tbank-ru
 
 # Tochka bank
 tochka-tech.com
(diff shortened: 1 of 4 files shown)


yetanotherqubick (changes requested):
1. Yet again I don't think it's necessary to add comments about `sber` and `tbank-ru` since these entries are self-explanatory.
2. I don't think it's a good idea to include whole `sber` into `category-travel-ru`. Perhaps it would be better to simply add a domain. It's Otello by the way.
Honestly I don't see much sense in `category-travel-ru` right now. Who and why would use it?.. Meanwhile it makes maintenance more complicated. We can't add domains with attributes and affiliations would break downstream as said by @MkQtS...
4. `my.com` is more than a email service. In its ToS it is referred as `MGL MY.COM (CYPRUS) LIMITED`.

MkQtS:
If there's few domains (only `otello.ru`?) that should be added into `category-travel-ru`, you can just add the domain directly in the category.

ACherepkov1989:
[yetanotherqubick](https://github.com/yetanotherqubick)
1. Made for convenience and understanding.
2. Yes, including the entire service is not the best idea, but several domains are interconnected there.
3. Corrected.

[MkQtS](https://github.com/MkQtS)
Authorization in the service is carried out via a phone number or id.sber.ru. The service also uses maps from 2gis.ru.

yetanotherqubick:
> Authorization in the service is carried out via a phone number or id.sber.ru. The service also uses maps from 2gis.ru.

See https://github.com/v2fly/domain-list-community/issues/3414#issuecomment-4231687489 by @MkQtS

MkQtS:
Adding additional authorization and map domains into the category seems fine, as it's generally reasonable, and not likely to break anything for the actual usage or user's expectation. Since the author prefer to add them, we can keep them.

MkQtS (approved):

PR #3477, opened 2026-04-15
Add Russian MyOffice

ACherepkov1989:


yetanotherqubick (commented):
I think a better name for the new category would be `myoffice-ru` according to similar cases with `-cn`, `-ir` or `-ru`.

ACherepkov1989:
> I think a better name for the new category would be `myoffice-ru` according to similar cases with `-cn`, `-ir` or `-ru`.

Maybe that's right. Found here "hsbc - hsbc-cn" and "jetbrains - jetbrains-ai".

yetanotherqubick (approved):

MkQtS (approved):

PR #3478, opened 2026-04-15
Create ucoz

ChaCha20-Poly-1305:
uCoz is a Russian SaaS web hosting. The same legal entity owns uKit website builder. There are many websites with its subdomains.

Domains are sourced from search engines and verified with BGP toolkit for the presence of uCoz NS, MX records. I've added it to `category:companies`, but I'm not sure if it should also go in `category-ru`.

### DIFF
--- a/data/category-companies
+++ b/data/category-companies
@@ -92,6 +92,7 @@ include:theboringcompany
 include:tonec
 include:uber
 include:ubiquiti
+include:ucoz
 include:verisign
 include:verizon
 include:vultr
--- a/data/category-ru
+++ b/data/category-ru
@@ -17,6 +17,7 @@ include:kaspersky
 include:mailru-group
 include:mindbox
 include:selectel
+include:ucoz-ru
 include:whoosh
 include:yandex
 
--- a/data/ucoz
+++ b/data/ucoz
@@ -1,0 +1,39 @@
+# ucoz
+
+include:ucoz-ru
+ucoz.ae
+ucoz.club
+ucoz.co
+ucoz.co.il
+ucoz.co.uk
+ucoz.com
+ucoz.com.br
+ucoz.com.ro
+ucoz.de
+ucoz.es
+ucoz.fr
+ucoz.hu
+ucoz.kz
+ucoz.lv
+ucoz.md
+ucoz.net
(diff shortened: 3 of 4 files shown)


yetanotherqubick:
Could you also add `ucoz` to `category-ru`? Thanks :)

ChaCha20-Poly-1305:
> Could you also add `ucoz` to `category-ru`? Thanks :)

Done.

yetanotherqubick (changes requested):
Hmm, now that I see UA domain zone here I think it shouldn't be included into `category-ru` as a whole... At least because UA TLD is blocked in Russia entirely. Perhaps it would be better to create a separate `ucoz-ru` with Russian domains only (`ucoz.ru`, `ucozmedia.ru`, `scripts-for-ucoz.ru`, `uguide.ru`)  and include that one in `category-ru` and `ucoz` itself.

ChaCha20-Poly-1305:
> Hmm, now that I see UA domain zone here I think it shouldn't be included into `category-ru` as a whole... At least because UA TLD is blocked in Russia entirely. Perhaps it would be better to create a separate `ucoz-ru` with Russian domains only (`ucoz.ru`, `ucozmedia.ru`, `scripts-for-ucoz.ru`, `uguide.ru`) and include that one in `category-ru` and `ucoz` itself.

Done. But in that case you may also want to do something about .ua domains present in `yandex` and `2gis`.

yetanotherqubick (approved):

yetanotherqubick:
Thanks once again!

> > Hmm, now that I see UA domain zone here I think it shouldn't be included into `category-ru` as a whole... At least because UA TLD is blocked in Russia entirely. Perhaps it would be better to create a separate `ucoz-ru` with Russian domains only (`ucoz.ru`, `ucozmedia.ru`, `scripts-for-ucoz.ru`, `uguide.ru`) and include that one in `category-ru` and `ucoz` itself.
> 
> Done. But in that case you may also want to do something about .ua domains present in `yandex` and `2gis`.

`yandex.ua` belongs to Yandex, but it doesn't seem to function. `2gis.ua` should be removed for sure.

MkQtS (approved):

PR #3481, opened 2026-04-15
Remove domain prefix for several Russian categories, small clean ups

yetanotherqubick:
Removed domain prefixes as a follow-up for #3464. There are also a few minor additions and deletions.

### DIFF
--- a/data/2gis
+++ b/data/2gis
@@ -11,5 +11,4 @@
 2gis.kz
 2gis.ru
 2gis.tj
-2gis.ua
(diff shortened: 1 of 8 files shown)


yetanotherqubick:
Removed `2gis.ua` following suggestion by @ChaCha20-Poly-1305 in https://github.com/v2fly/domain-list-community/issues/3478#issuecomment-4256238459

MkQtS (approved):

PR #3482, opened 2026-04-16
category-ai-!cn: add minimax.io

arturbikbaev:


MkQtS:
It's serving global area, not cn. The domain does not resolve to a cn IP. So it should be added to `category-ai-!cn` instead.

MkQtS (approved):

PR #3466, opened 2026-04-13
category-ru: add include:category-media-ru

fullstackPrincess:
## Summary

Add `include:category-media-ru` to `category-ru`.

Currently, `category-ru` includes several Russian subcategories:
- `category-ecommerce-ru`
- `category-entertainment-ru`
- `category-gov-ru`
- `category-medicine-ru`
- `category-retail-ru`
- `category-travel-ru`

However, `category-media-ru` is missing. This means ~130 Russian media domains (meduza.io, novayagazeta.ru, tvrain.tv, mediazona.ca, holod.media, etc.) are not reachable via `geosite:category-ru`.

This PR adds the missing include so that Russian media outlets are part of the broader Russian domain routing category.

PR #3465, opened 2026-04-13
Add more Russian service domains

fullstackPrincess:
## Summary

Add 24 new Russian service domains distributed across 7 data files:

### `data/2gis`
- `2gis.tech` — technical/developer domain for 2GIS

### `data/category-retail-ru`
- `magnit.com` — corporate site for Magnit retail chain (alongside existing `magnit.ru`)

### `data/mailru-group`
- `uma.media` — United Media Agency, music distributor owned by VK Group

### `data/category-media` (near existing `rt.com`)
- `russiatoday.com` — former domain of RT (Russia Today)
- `rteam.tech` — IT company owned by ANO TV-Novosti (RT parent org)
- `shop-rt.com` — RT merchandise store
- `swentr.site` — RT mirror domain

### `data/category-media-ru`
- `3dnews.kz` — 3DNews tech media (Kazakh mirror)
- `tsargrad.tv` — Tsargrad TV channel

### `data/category-entertainment-ru`
- `promodj.com` — electronic music and DJ community platform

### `data/category-ru`
- `gismeteo.st`, `gismeteo.net` — Gismeteo weather service (CDN and API)
- `kontur.host` — SKB Kontur, business SaaS provider
- `envybox.io` @ads, `whitesaas.com` @ads, `saas-support.com` @ads — Envybox conversion widgets
- `flomni.com` — omnichannel communications platform
- `3ebra.net` — Zebra Net Bureau, web development agency
- `flocktory.com` @ads — marketing personalization platform
- `targetads.io` @ads — advertising verification platform
- `kinescopecdn.net` — Kinescope video hosting CDN
- `yourgood.app` — CRM for small business
- `full:ru.id.group-ib.com` — Group-IB Russian auth portal
- `full:lk.ognihome.com` — OGNI HOME resident portal

**Evidence**: All domains are verified as Russian services (hosted on Russian infrastructure, registered via Russian registrars, or owned by Russian companies).

PR #3486, opened 2026-04-16
Fix for category-media-ru and category-media-ru-blocked

MkQtS:
Link to #2622, #3465, #3466


MkQtS:
@DanielLavrushin @fullstackPrincess @yetanotherqubick @ACherepkov1989 

Feel free to share your ideas.

ghost:
why the list cannot have `@tag` support, for example:

```
# This filename is "category-media", contains ALL media websites
# new tag syntax: @cc:<CC>, where <CC> is lowecased 2letter country code
# @cc:!<CC> means "not <CC>"

ok-in-russia.com
blocked-in-russia.com @cc:!ru
blocked-in-china-and-russia.com @cc:!cn @cc:!ru
only-accessible-from-russia.ru @cc:ru
for-japanese-people-only.jp @cc:jp
advert-russia.com @ads @cc:ru 
```

MkQtS:
> why the list cannot have `@tag` support, for example:
> 
> ```
> # This filename is "category-media", contains ALL media websites
> # new tag syntax: @cc:<CC>, where <CC> is lowecased 2letter country code
> # @cc:!<CC> means "not <CC>"
> 
> ok-in-russia.com
> blocked-in-russia.com @cc:!ru
> blocked-in-china-and-russia.com @cc:!cn @cc:!ru
> only-accessible-from-russia.ru @cc:ru
> for-japanese-people-only.jp @cc:jp
> advert-russia.com @ads @cc:ru 
> ```

This may cause unpredictable results in downstream projects/clients. And end users are not likely to learn it.

yetanotherqubick (approved):
I'm a bit skeptical about `category-media-!ru` name, but it's probably the least worst option

t0ximaxi:
`category-media-ru` and `category-media-ru-forbidden` maybe?

ghost:
Current `!cc` names are not expained clearly. If this means `this list is not accessible in CC`, then how about:

- `category-media_ru-blocked`
- `category-media_not-ru`
- `category-media_NOTru`

ACherepkov1989:
@MkQtS, this solution is good, similar to the existing !cn.
@sranders1, suggested the idea of `category-media-ru-blocked` and it is better suited for user understanding.

MkQtS:
I've renamed it into `category-media-ru-blocked`. There seem to be no objections. Let's merge this.

vluisteren:
> I've renamed it into `category-media-ru-blocked`. There seem to be no objections. Let's merge this.

I don't think it keeps the same naming format like the the existing, can I rename it into `category-media-blocked-ru` the region code is always at the last. @MkQtS

vluisteren:
And I am confusing about why it contains the many `.ru` TLD domains. Isn't the `.ru` TLD managed by the Putin regime?

The registry office can just delete them?

MkQtS:
> > I've renamed it into `category-media-ru-blocked`. There seem to be no objections. Let's merge this.
> 
> I don't think it keeps the same naming format like the the existing, can I rename it into `category-media-blocked-ru` the region code is always at the last. @MkQtS

It may break existing configurations. So please not. @vluisteren

vluisteren:
> > > I've renamed it into `category-media-ru-blocked`. There seem to be no objections. Let's merge this.
> > 
> > 
> > I don't think it keeps the same naming format like the the existing, can I rename it into `category-media-blocked-ru` the region code is always at the last. @MkQtS
> 
> It may break existing configurations. So please not. @vluisteren

I see. I would just use my [normailization script](https://github.com/vluisteren/domain-list-community/blob/normalize/scripts/normalize_data_files.py) to knock it downstream.

But for the existing configuration, are you talking the V2ray series may implement a rule for that(Russian users)? Or there is something else I would take care of it?

MkQtS:
> But for the existing configuration, are you talking the V2ray series may implement a rule for that(Russian users)? Or there is something else I would take care of it?

Configurations of v2ray (and some other clients) users may contain certain existing geosite list. If we deleted/renamed an existing list, their clients fails and we get complaints. So we need to be cautious and only do so when it's safe/necessary. 

There seems no written rules for this. This is more a personal opinion to keep things stable.

As you can see, we have few active mantainers/contributors. And current structure works. Systematic (breaking) changes are not likely to happen without enough needs and manpower.

vluisteren:
> > But for the existing configuration, are you talking the V2ray series may implement a rule for that(Russian users)? Or there is something else I would take care of it?
> 
> Configurations of v2ray (and some other clients) users may contain certain existing geosite list. If we deleted/renamed an existing list, their clients fails and we get complaints. So we need to be cautious and only do so when it's safe/necessary.
> 
> There seems no written rules for this. This is more a personal opinion to keep things stable.
> 
> As you can see, we have few active mantainers/contributors. And current structure works. Systematic (breaking) changes are not likely to happen without enough needs and manpower.

There should be an agreement on how to fix the error, or the errata would be unmanageable. Like we can notify the downstream for several week, they can ship those changes in the next update.

MkQtS:
I don't treat these naming/calssfication issues as errors, which must be fixed. Making breaking changes and get limited benefits, this is unrealistic.

It's a doubtable that whether the downstream would agree/adapt to these changes. Moreover, there're end users who are using old/stable configurations. It would be a long and maybe painful process.

Talks here are kind of abstract. Actually, you may make a fork, do changes as you want, and show the advantages to persuade users. With enough support, we can make changes.

PR #3516, opened 2026-04-24
Add more cn domains

refear99:


EpLiar:
with dig cli's result, `dingxx.com`'s A record isn't in China mainland
```
bash-3.2$ dig A +short dingxx.com 223.5.5.5
156.238.255.22
```
image

PR #3530, opened 2026-04-29
Update category-public-tracker

TripleA150:
Blocking by subdomains is ineffective due to their frequent changes.

I have added “full” to certain domains that are used not only as trackers


TripleA150:
@MkQtS and can you review this commit, thanks!

MkQtS:
The change is significant. I'm not sure if this is appropriate.

ghost:
Removing `full:` will include all subdomains, which can lead to false positive. Please don't do this or re-verify what you're doing.

TripleA150:
> Removing `full:` will include all subdomains, which can lead to false positive. Please don't do this or re-verify what you're doing.

it's already done

PR #3533, opened 2026-04-30
Update category-ip-geo-detect

sngvy:


EpLiar:
I checked this domain, it's OK but you should sort it.

sngvy:
I've sorted and placed the domain in the right place. Check it out again.

EpLiar (approved):

PR #3541, opened 2026-05-05
Russian bank update

TripleA150:


yetanotherqubick (changes requested):
1. Raiffeisen is a group of European banks and shouldn't be included in `category-bank-ru` entirely. Instead I suggest to create `raiffeisen-ru` and include it both in `raiffeisen` and `category-bank-ru`.
2. Source for `sber.world`, `sber.ws` and `sbercoin.com` relationship to Sber?
3. Why did you create `Other` section, while there're sections for every bank?
3.1. MTS already has its own section, but is also included there. 
4. `autobahn.db.com` doesn't seem related to Russian banking entirely. Same for `mizuhogroup.com`.

TripleA150:
@yetanotherqubick
1. It's not necessary to create another list for 2 domains. Will delete category and add only russian part
2. Will delete sber.world and sber.ws. sbercoin.com is related to Sber, just open the link or yandex it :)
3. "Other" section is for single domains and not popular banks
3.1. Will be combined
4. Deutsche Bank - https://cbr.ru/finorg/foinfo/?ogrn=1027739369041. Will delete autobahn.db.com and add russian domains
Mizuho Bank https://www.cbr.ru/finorg/foinfo/?ogrn=1027700317028

yetanotherqubick (changes requested):
1. It's actually perfectly fine to create both `raiffeisen` and `raiffeisen-ru`. For example, there are already `ucoz` and `ucoz-ru`. Would be better, if we could use `@ru` attribute similar to `@cn`, but maintainers won't allow it for some reasons.
2. I still wasn't able to link `sbercoin.com` to Sber. It's resolved to Cloudflare IPs, is registered by some resident of Astana, Kazakhstan, who is using ProtonMail, and is even flagged by Yandex Safebrowsing as phishing. Doesn't sound good to me.
3. If Mizuho Bank doesn't have Russian website, its international website shouldn't be included, since it's not Russian entity. However, Mizuho Bank does have Russian domain, which can be found on its website: `mhcbr.ru`. `mhcb.ru` also belongs to them.

yetanotherqubick:
@MkQtS why can't we use `@ru` attribute, by the way? Is it better to create separate files in these cases, like with `ucoz`/`ucoz-ru` and `raiffeisen`/`raiffeisen-ru`?

TripleA150:
@yetanotherqubick 
1. I actually can't support your idea. There're no logical reasons to divide Russian Raiffeisen domains into a new list. You can make your own PR if you really think it needs to be
2,3. Will be changed

MkQtS:
> Would be better, if we could use `@ru` attribute similar to `@cn`, but maintainers won't allow it for some reasons.

> @MkQtS why can't we use `@ru` attribute, by the way?

`@ru` is reasonable for me. It should work fine here in this repo or `*ray`. But we'd better asking for other maintainers to avoid causing unforeseen issues for downstream projects/end users.

> Is it better to create separate files in these cases, like with `ucoz`/`ucoz-ru` and `raiffeisen`/`raiffeisen-ru`?

I don't know much about these companies. Generally it is acceptable to create separate files. It's up to you.

TripleA150:
@MkQtS At this time there are no changes I want to add. I prefer to divide the list by using @ru soon.

yetanotherqubick:
> @MkQtS At this time there are no changes I want to add. I prefer to divide the list by using @ru soon.

Can you add comments on Raiffeisen and OTP at least? Not to forget about improving that part in the future

TripleA150:
@yetanotherqubick @MkQtS Done

yetanotherqubick (approved):

TripleA150:
@MkQtS any updates?

MkQtS:
I've been a bit busy lately...

MkQtS (approved):

PR #3292, opened 2026-02-19
Google Gemini adjustments

Sharm:
Google Gemini use a lot of new *.clients6.google.com domains

PR #3542, opened 2026-05-06
Update github & github-copilot

KukiSa:
The regex matching for `^github-production-release-asset-[0-9a-zA-Z]{6}\.s3\.amazonaws\.com$` is inaccurate; GitHub does not utilize any domains matching this pattern other than `github-production-release-asset-2e65be.s3.amazonaws.com`. Since Amazon S3 bucket naming requires no verification, an overly broad matching range could lead to unintended behavior.

The same logic applies to the matching of `blob.core.windows.net`, as all Azure Blob Storage accounts share the same domain suffix.

### DIFF
--- a/data/github
+++ b/data/github
@@ -26,13 +26,37 @@ opensource.guide
 repo.new
 thegithubshop.com
 
-# GitHub Actions Artifacts
-blob.core.windows.net
-
 full:github-api.arkoselabs.com
+
+# Amazon S3
 full:github-cloud.s3.amazonaws.com
-regexp:^github-production-release-asset-[0-9a-zA-Z]{6}\.s3\.amazonaws\.com$
+full:github-production-release-asset-2e65be.s3.amazonaws.com
+full:github-production-repository-file-5c1aeb.s3.amazonaws.com
+full:github-production-repository-image-32fea6.s3.amazonaws.com
+full:github-production-upload-manifest-file-7fdce7.s3.amazonaws.com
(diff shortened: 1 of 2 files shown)


k-g-a:
Hi! 
Thanks for the contribution.

I was inspecting network logs in search of "why copilot does not work with geosite:github-copilot?" and found some more domains to add (in addition to those you've already added):

- `domain:copilot.microsoft.com`
- `full:copilot-telemetry-service.githubusercontent.com`
- `full:origin-tracker.githubusercontent.com`

But the most important is that IDE copilot plugins (at least for jetbrains products) judge your origin (and allowence) by the `api.github.com` call (at least as of april-may 2026), meaning that we need to add `full:api.github.com` for `github-copilot` rule to be usable for development purposes.

MkQtS:
> why copilot does not work with geosite:github-copilot?

@k-g-a, using `geosite:github` is better than `geosite:github-copilot` for your purpose here.

> we need to add full:api.github.com for github-copilot rule to be usable for development purposes.

It's safe to add *exclusive* domains to a certain list. But adding generic domains to a dedicated list may break users' expectation and cause unpredictable issues, similar case in #3292. If it's really necessary, please open another PR.

As tech companies are tending to use some generic domains all over their business, users can use a broader list, like `geosite:github` for GitHub Copilot and `geosite:google` for Google Gemini, to ensure that traffic appears normal.

---

I'll add `copilot-telemetry-service.githubusercontent.com` to `github-copilot`.

MkQtS (approved):

PR #3552, opened 2026-05-09
Add tb.ewintech.cn to cn list

GGboy1994:
Add my domain to the cn list for direct access to my file management system.

### DIFF
--- a/data/cn
+++ b/data/cn
@@ -2,3 +2,4 @@
 
 include:tld-cn
 include:geolocation-cn
+tb.ewintech.cn


MkQtS:
There is `[domain:]cn` in `data/tld-cn` which matches all `*.cn` domains. It's useless to add a certain `*.cn` domain into `data/cn` because it's redundant for the program.

You can add it to `geolocation-cn`, in a proper section. BTW, why not you add `ewintech.cn` instead of `tb.ewintech.cn`?

PR #3556, opened 2026-05-09
category-ip-geo-detect: add new IP geolocation detection services

jinndi:
Add beacondb.net, ipbot.com, and iplookup.it to the IP geolocation detection category.

### DIFF
--- a/data/category-ip-geo-detect
+++ b/data/category-ip-geo-detect
@@ -7,7 +7,6 @@ include:ipip
 abstractapi.com
 apiip.net
 apivoid.com
-beacondb.net
 bigdatacloud.net
 check-host.net
 checkip.org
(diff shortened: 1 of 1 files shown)


yetanotherqubick:
Does `beacondb.net` really belong there? Its wireless geolocation database and other domains provide IP address lookup functionality.

jinndi:
> Does `beacondb.net` really belong there? Its wireless geolocation database and other domains provide IP address lookup functionality.

Fixed to `full:api.beacondb.net`

It detects my server IP in the IP leak detection program.

jinndi:
Actually no, I tested it without this domain and nothing like that happened, so I removed it. The other two are needed.

MkQtS (approved):

PR #3557, opened 2026-05-09
Create category-ads-ru

yetanotherqubick:
Created `category-ads-ru` and included it in `category-ru` and `category-ads-all`. I also moved `category-ads-ir` from `category-ads` to `category-ads-all`.

Source for domains: VirusTotal passive DNS replication history.

### DIFF
--- a/data/category-ads
+++ b/data/category-ads
@@ -1,5 +1,4 @@
 # This file contains domains that clearly serving ads
-include:category-ads-ir
 
 include:acfun @ads
 include:adobe @ads
--- a/data/category-ads
+++ b/data/category-ads
@@ -1,5 +1,7 @@
 # This file contains domains that clearly serving ads
 
+include:category-ads-ir
+
 include:acfun @ads
 include:adobe @ads
 include:alibaba @ads
--- a/data/category-ads
+++ b/data/category-ads
@@ -1,6 +1,7 @@
 # This file contains domains that clearly serving ads
 
 include:category-ads-ir
+include:category-ads-ru
 
 include:acfun @ads
 include:adobe @ads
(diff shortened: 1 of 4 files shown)


ACherepkov1989:
I think it's wrong to move `include:carrotquest`, `include:envybox`, and `include:mindbox` entirely from `category-ru` to `category-ads-ru`. The solution you proposed is as follows: when a user uses category-ads-ru to block ads and trackers, the main domains in these lists will also be blocked, making it impossible to even access the `carrotquest`, `envybox`, and `mindbox` websites.
The correct solution would be to simply add these companies with the `@ads` prefix to `category-ads-ru` without removing them from `category-ru`.

yetanotherqubick:
`category-ads-ru` itself is included in `category-ads-all` and `category-ru`. If one doesn't want to block ads providers domains (only ads domains explicitly), then they should use `category-ads`. If that doesn't matter to them, they could use `category-ads-all`. I don't think there're any problems.

Of course, someone *could* use `category-ads-ru` separately, but I think it's *much* less likely than someone using `category-ads` or `category-ads-all`. To do that you must know this category exists in the first place. To find out about it you'd need to open this repository and look through existing categories. Can someone use that category without understanding what's inside it? Sure. Who's responsibility it is? I say it's on the end user.

I will add a notice on that. Should be fine I think.

MkQtS:
Any reason to move `category-ads-ir`? Will this affect Iranian users?

ACherepkov1989:
> Of course, someone _could_ use `category-ads-ru` separately, but I think it's _much_ less likely than someone using `category-ads` or `category-ads-all`. To do that you must know this category exists in the first place. To find out about it you'd need to open this repository and look through existing categories. Can someone use that category without understanding what's inside it? Sure. Who's responsibility it is? I say it's on the end user.

I believe that this position is incorrect.

yetanotherqubick:
> Any reason to move `category-ads-ir`? Will this affect Iranian users?

Double-checked it just now — yes, it should stay in `category-ads`.

I think I'll also change `category-ads-ru` so it only lists explicitly advertising domains.

PR #3592, opened 2026-05-16
geolocation-!cn: add zhangzhehan.net

LoveLemonOnFish:
Add geosite category for zhangzhehan.net

The website appears inaccessible from mainland China without proxy/VPN. Multiple users observed connectivity issues including DNS/TLS anomalies from mainland China networks, while the site remains accessible outside China.

Adding this domain may help users configure proxy routing rules more easily.

### DIFF
--- a/data/geolocation-!cn
+++ b/data/geolocation-!cn
@@ -339,3 +339,4 @@ include:xdty
 include:xingrz
 
 sinyalee.com
+zhangzhehan.net


MkQtS:
1. The `domain` type (default) matches subdomains. `[domain:]zhangzhehan.net` matches `[domain:]www.zhangzhehan.net`, so the later one is duplicate
2. The domain is not added/included in any other lists. Unless users explicitly uses `geosite:zhangzhehan` in their config, current changes have no practical effect.

The proper way is to add `zhangzhehan.net` in `data/geolocation-!cn`

MkQtS (approved):

PR #3604, opened 2026-05-19
Create vokino and add ru Mobile IMS & Banking Services

steepenergy:
add Mobile IMS
sources:
https://denis.elib.ru/vowifi/
https://www.mcc-mnc.net/

add
E-commerce acquiring
Retail Banking Services Gate
Retail Banking Services User Acceptance Testing

add
Reaviz to category-medicine-ru

add Vokino (Online Cinema)

### DIFF
--- a/data/category-ru
+++ b/data/category-ru
@@ -81,6 +81,7 @@ include:t2-ru
 adygtelecom.com    # Adygea Telephone Company, regional operator
 beeline.ru
 dom.ru
+mcc250.pub.3gppnetwork.org # Mobile IMS
 
 # Other services
 3ebra.net          # Zebra Net Bureau, web development agency
--- a/data/category-ru
+++ b/data/category-ru
@@ -65,6 +65,7 @@ include:rostelecom
 include:t2-ru
 beeline.ru
 dom.ru
+mcc250.pub.3gppnetwork.org #VoWiFi (mcc250 - Mobile Country Code Russian Federation)
 
 # Other services
 3ebra.net          # Zebra Net Bureau, web development agency
--- a/data/category-ru
+++ b/data/category-ru
@@ -65,7 +65,7 @@ include:rostelecom
 include:t2-ru
 beeline.ru
 dom.ru
-mcc250.pub.3gppnetwork.org # VoWiFi (mcc250 - Mobile Country Code Russian Federation)
+mcc250.pub.3gppnetwork.org # Mobile IMS
 
 # Other services
 3ebra.net          # Zebra Net Bureau, web development agency
--- a/data/category-ru
+++ b/data/category-ru
@@ -65,7 +65,7 @@ include:rostelecom
 include:t2-ru
 beeline.ru
 dom.ru
-mcc250.pub.3gppnetwork.org #VoWiFi (mcc250 - Mobile Country Code Russian Federation)
+mcc250.pub.3gppnetwork.org # VoWiFi (mcc250 - Mobile Country Code Russian Federation)
 
 # Other services
 3ebra.net          # Zebra Net Bureau, web development agency
--- a/data/category-bank-ru
+++ b/data/category-bank-ru
@@ -104,6 +104,7 @@ bank-arzamas.ru
 bank-credit-suisse-moscow.ru
 bank131.com
 bankermak.ru
+best2pay.net
 bfa.ru
 blanc.ru
 bm-bank.ru
@@ -112,6 +113,7 @@ commerzbank.ru
 finambank.ru
 forshtadt.ru
 homecredit.ru
+hva.im # Sovcombank Halva
 ibamoscow.ru
 jpmorgan.ru
 korona.com
@@ -123,6 +125,8 @@ primbank.ru
 prvbank.ru
 pskb.ru
 pvubank.com
+rbsgate.com
+rbsuat.com
 rencapbank.ru
 rencredit.ru
 rncoluminis.ru
--- a/data/category-bank-ru
+++ b/data/category-bank-ru
@@ -109,6 +109,7 @@ commerzbank.ru
 finambank.ru
 forshtadt.ru
 homecredit.ru
+hva.im # Совкомбанк Халва
 ibamoscow.ru
 jpmorgan.ru
 korona.com
--- a/data/category-bank-ru
+++ b/data/category-bank-ru
@@ -101,6 +101,7 @@ bank-arzamas.ru
 bank-credit-suisse-moscow.ru
 bank131.com
 bankermak.ru
+best2pay.net
 bfa.ru
 bm-bank.ru
 bspb.ru
@@ -119,6 +120,8 @@ primbank.ru
 prvbank.ru
 pskb.ru
 pvubank.com
+rbsgate.com
+rbsuat.com
 rencapbank.ru
 rencredit.ru
 rncoluminis.ru
--- a/data/category-bank-ru
+++ b/data/category-bank-ru
@@ -109,7 +109,7 @@ commerzbank.ru
 finambank.ru
 forshtadt.ru
 homecredit.ru
-hva.im # Совкомбанк Халва
+hva.im # Sovcombank Halva
 ibamoscow.ru
 jpmorgan.ru
 korona.com
--- a/data/category-medicine-ru
+++ b/data/category-medicine-ru
@@ -36,5 +36,9 @@ polyclinika.ru
 ## Dentistry Zub Ru – network of dental clinics in Moscow
 zub.ru
 
+# Reaviz
+reaviz.com
+reaviz.ru
+
 # St. Petersburg city polyclinics and medical institutions
 gorzdrav.spb.ru
--- a/data/category-medicine-ru
+++ b/data/category-medicine-ru
@@ -19,5 +19,9 @@ lk.emias.mos.ru
 # Polyclinika-ru
 polyclinika.ru
 
+# Reaviz
+reaviz.com
+reaviz.ru
+
 # Vipmed
 vipmed.ru
--- a/data/vokino
+++ b/data/vokino
@@ -1,0 +1,25 @@
+# Vokino (Online Cinema)
+vokino.org
+vokino.pro
+vokino.tv
+
+# CDNs & balancers
+85b6b2.ws
+ashdi.vip
+cdn2cdn.com
+cdn2site.com
+cdnsqu.com
+cdntogo.net
+digital-cdn.net
+fotpro135alto.com
+interkh.com
+loadbox.ws
+showvid.ws
+rstprgapipt.com
+rtbcdn.cloud
+srvkp.com
+stream-balancer-allo-1.live
+stream-balancer-allo-1.site
+videobase.biz
+vkvideo.cloud
+werkecdn.me
--- a/data/vokino
+++ b/data/vokino
@@ -4,15 +4,20 @@ vokino.pro
 vokino.tv
 
 # CDNs & balancers
+ashdi.vip
+cdn.videobase.biz
 cdn2cdn.com
 cdn2site.com
 cdnsqu.com
 cdntogo.net
+cloud.mail.ru
 digital-cdn.net
 fotpro135alto.com
 interkh.com
+kinescopecdn.net
 loadbox.ws
 rstprgapipt.com
+rtbcdn.cloud
 stream-balancer-allo-1.live
 stream-balancer-allo-1.site
 videobase.biz
--- a/data/vokino
+++ b/data/vokino
@@ -4,6 +4,7 @@ vokino.pro
 vokino.tv
 
 # CDNs & balancers
+85b6b2.ws
 ashdi.vip
 cdn2cdn.com
 cdn2site.com
@@ -14,6 +15,7 @@ fotpro135alto.com
 interkh.com
 kinescopecdn.net
 loadbox.ws
+showvid.ws
 rstprgapipt.com
 rtbcdn.cloud
 srvkp.com
--- a/data/vokino
+++ b/data/vokino
@@ -10,7 +10,6 @@ cdn2cdn.com
 cdn2site.com
 cdnsqu.com
 cdntogo.net
-cloud.mail.ru
 digital-cdn.net
 fotpro135alto.com
 interkh.com
--- a/data/vokino
+++ b/data/vokino
@@ -1,0 +1,19 @@
+# Vokino (Online Cinema)
+vokino.org
+vokino.pro
+vokino.tv
+
+# CDNs & balancers
+cdn2cdn.com
+cdn2site.com
+cdnsqu.com
+cdntogo.net
+digital-cdn.net
+fotpro135alto.com
+interkh.com
+loadbox.ws
+rstprgapipt.com
+stream-balancer-allo-1.live
+stream-balancer-allo-1.site
+videobase.biz
+werkecdn.me
--- a/data/vokino
+++ b/data/vokino
@@ -5,7 +5,6 @@ vokino.tv
 
 # CDNs & balancers
 ashdi.vip
-cdn.videobase.biz
 cdn2cdn.com
 cdn2site.com
 cdnsqu.com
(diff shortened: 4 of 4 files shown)


ACherepkov1989:
The `mcc250.pub.3gppnetwork.org` domain is related to the `IMS` architecture (a solution for implementing services in IP-based communication networks).
IMS-based technologies include VoLTE (Voice over LTE), VoWiFi (Voice over Wi-Fi) and ViLTE (Video over LTE).
Therefore `# VoWiFi` comment is not correct and should be `# Mobile IMS`.

steepenergy:
> The `mcc250.pub.3gppnetwork.org` domain is related to the `IMS` architecture (a solution for implementing services in IP-based communication networks). IMS-based technologies include VoLTE (Voice over LTE), VoWiFi (Voice over Wi-Fi) and ViLTE (Video over LTE). Therefore `# VoWiFi` comment is not correct and should be `# Mobile IMS`.

Thank you for the detailed explanation! You are absolutely right, that is much more accurate. I have changed the comment to `# Mobile IMS`

steepenergy:
@ACherepkov1989 Hi, could you please take another look?

steepenergy:
@ACherepkov1989 ?

ACherepkov1989:
> Hi, could you please take another look?

You need to wait for a merge decision from the project developer...

steepenergy:
Hi everyone! Just a gentle reminder to see if anyone has time to review this change. Thanks!

data/category-ru
```diff
@@ -80,6 +80,7 @@ include:t2-ru
 adygtelecom.com    # Adygea Telephone Company, regional operator
 beeline.ru
 dom.ru
+mcc250.pub.3gppnetwork.org # Mobile IMS
```
MkQtS, line 84: Why this domains belong to ru?

steepenergy: Why this domain is needed in category-ru:

The domain mcc250.pub.3gppnetwork.org is strictly tied to Russia (MCC 250 = Russian Federation). It is used by smartphones to resolve the Evolved Packet Data Gateway (ePDG) and IMS servers for Russian mobile operators (e.g., MTS, Megafon) to establish VoWiFi (Wi-Fi Calling) and VoLTE connections

The issue without this rule:
If this domain is not categorized as RU, proxy clients (like v2ray/Xray/sing-box) will route DNS queries and subsequent traffic to this domain through a foreign proxy server

Russian mobile operators strictly geo-block their ePDG gateways. If the IPsec tunnel request (UDP 500/4500) comes from a foreign IP address rather than a native Russian IP, the operator drops the connection. This completely breaks Wi-Fi Calling for users inside Russia when the proxy is active

The solution:
By adding mcc250.pub.3gppnetwork.org to category-ru, it triggers standard direct routing rules for Russian traffic. This ensures the smartphone connects to the local mobile operator's gateway using the user's real Russian IP, keeping Wi-Fi Calling functional

steepenergy: sources:
https://denis.elib.ru/vowifi/
https://www.mcc-mnc.net/

MkQtS:
1. Please include vokino in relevant category
2. Please check these duplicate domains (already in `data/` but readded in this PR)

```
./data/category-medicine-ru
10:vipmed.ru
./data/kinopub
19:cdn2cdn.com
./data/kinopub
20:cdn2site.com
./data/category-ru
91:kinescopecdn.net   # Kinescope video hosting CDN
```

steepenergy:
@MkQtS 
Hi! Thanks for the review. I have updated the PR with a structural improvement to avoid routing conflicts.

Here is what I did:

1.  **Created `data/category-entertainment-ru-blocked`**: I couldn't add `vokino` to the existing `category-entertainment-ru` because that list contains legal RU streaming services (like Kinopoisk) that actively block VPNs and **must** be routed `DIRECT`. Vokino and Kinopub, on the other hand, are unofficial services blocked in Russia and **must** be routed via `PROXY`. I created a new category to logically group them for users.
    
2.  **Regarding the duplicates**: I removed `kinescopecdn.net` from my file, as it is a legitimate B2B CDN already covered by `category-ru`. However, I **kept** `cdn2cdn.com` and `cdn2site.com` in `data/vokino`. Even though they are duplicated in `kinopub`, removing them from `vokino` would mean the standalone rule won't work properly. Any future user who configures their proxy to route strictly `geosite:vokino` would have a broken video stream. Since the compiler automatically removes duplicates during the `geosite.dat` build process anyway, keeping them in the source files ensures that standalone routing functions correctly for both services without bloating the final compiled list.
    

Please let me know if this approach is approved!

MkQtS:
@steepenergy 

Thanks for your detailed explanation.

For the new list, the name `category-entertainment-ru-blocked` may not conform to our naming convention, as @vluisteren pointed out in #3486. Since there're only two lists included in, maybe we can postpone the creation of the new category list, and wait for broader feedback from the community.

steepenergy:
@MkQtS 
Thanks for pointing that out. I read through #3486 and completely understand the naming convention issue (keeping `-ru` at the end) as well as your concern about creating a category for just two lists.

I found another similar service (`hdrezka`) that is already in the repository. So, to make the category truly useful and compliant, I did the following:

1.  Created `data/category-entertainment-unofficial-ru`. This follows the naming convention (keeping the `-ru` suffix at the end) and avoids the debated `blocked` keyword.
    
2.  Included `hdrezka`, `kinopub`, and the new `vokino` in this category.
    

Now we have three major services, which makes it a proper, solid grouping for the unofficial RU streaming ecosystem that users *must* route via PROXY.

(I also kept the shared CDNs in `data/vokino` so standalone routing still works, as the compiler handles duplicates anyway).

**If you still prefer not to introduce a new category right now, just let me know. I am completely fine with removing the category file and leaving only the standalone `vokino` list so we can proceed.**

Does this expanded category look good to you, or should we go with just the standalone file?

vluisteren:
> @MkQtS Thanks for pointing that out. I read through #3486 and completely understand the naming convention issue (keeping `-ru` at the end) as well as your concern about creating a category for just two lists.
> 
> I found another similar service (`hdrezka`) that is already in the repository. So, to make the category truly useful and compliant, I did the following:
> 
> 1. Created `data/category-entertainment-unofficial-ru`. This follows the naming convention (keeping the `-ru` suffix at the end) and avoids the debated `blocked` keyword.

I was thinking whether we should have a `geolocation-!ru`? 

The problem about the category is, we can't reuse them among every countries, every countries use a different name for the same thing.
https://github.com/v2fly/domain-list-community/pull/3857  is one I am trying to fix. 

Also, a few category only has a domain (ex. https://github.com/v2fly/domain-list-community/blob/master/data/category-outsource-cn ), and its name is confusing (A more clear name could be talnet marketplace or on demand service marketplace)

> 2. Included `hdrezka`, `kinopub`, and the new `vokino` in this category.
> 
> Now we have three major services, which makes it a proper, solid grouping for the unofficial RU streaming ecosystem that users _must_ route via PROXY.
> 

Besides, I would like to know what happens to the white list (allow list) route configuration? Is not a default configuration to the users in Russian?

In that case, I think adding a domain to a place which not in Russian(without region code) should work.

> (I also kept the shared CDNs in `data/vokino` so standalone routing still works, as the compiler handles duplicates anyway).
> 
> **If you still prefer not to introduce a new category right now, just let me know. I am completely fine with removing the category file and leaving only the standalone `vokino` list so we can proceed.**
> 
> Does this expanded category look good to you, or should we go with just the standalone file?

steepenergy:
@vluisteren @MkQtS

You make excellent points, and I agree. I will drop the `data/category-entertainment-unofficial-ru` category for now because:

*   **Size:** There are only 3 services to group anyway.
    
*   **Regional aspect:** You are right that these streaming platforms are actually used not only in Russia but also in Kazakhstan, Ukraine, and Belarus.
    
*   **Routing concern (most importantly):** Categories ending with `-ru` are indeed heavily used in routing configurations as proxy exclusions (to route traffic `DIRECT`). Naming a proxy-required list with `-ru` at the end could easily confuse users and break their setups.
    

I have reverted the category creation and left only the standalone `data/vokino` file for configurations that use targeted proxy routing (keeping its shared CDNs inside so it works properly).

The PR is updated and ready. Thanks for the great feedback!

vluisteren:
> @vluisteren @MkQtS
> 
> You make excellent points, and I agree. I will drop the `data/category-entertainment-unofficial-ru` category for now because:
> 
> * **Size:** There are only 3 services to group anyway.
> * **Regional aspect:** You are right that these streaming platforms are actually used not only in Russia but also in Kazakhstan, Ukraine, and Belarus.
> * **Routing concern (most importantly):** Categories ending with `-ru` are indeed heavily used in routing configurations as proxy exclusions (to route traffic `DIRECT`). Naming a proxy-required list with `-ru` at the end could easily confuse users and break their setups.
> 
> I have reverted the category creation and left only the standalone `data/vokino` file for configurations that use targeted proxy routing (keeping its shared CDNs inside so it works properly).
> 
> The PR is updated and ready. Thanks for the great feedback!

I am not the maintainer, while I saw many commits are changes to the same files, would you mind squashing them together, it would look ever better if you keep a commit for each of file or purpose.

Besides, do you know why there are `.ru` domains at `data/category-media-ru-blocked` ?

Doesn't Putin just delete their records? I need a insider of Russian to help me here.

steepenergy:
Hi @vluisteren,

I have squashed all the commits into a single clean one as you suggested.

As for your question about `.ru` domains and Russian censorship, the Russian government (via Roskomnadzor) primarily blocks websites at the ISP level using DPI (Deep Packet Inspection) and a state-controlled DNS infrastructure called the **National Domain Name System (NSDI)**. They don't usually delete domain records from the registry.

When a site gets banned, its domain is added to a national blacklist. Local internet providers are forced to drop connections based on IP/SNI, and the default state DNS (NSDI) acts as a giant sinkhole, returning fake or empty responses.

Revoking or actually "deleting" a `.ru` domain globally is a completely different, slow legal process (usually reserved for malware/phishing).

Because of this, many blocked `.ru` domains remain perfectly active, delegated, and resolvable in the global DNS system. You just cannot establish a connection to them from inside Russia without a VPN or Proxy

MkQtS (approved):

MkQtS:
> I was thinking whether we should have a `geolocation-!ru`?

We may need some kind of consensus or standard before creating such a list.

*I don't live in Russia. These opinions may not reflect the actual situation*

To tell whether a domain is `cn` or `!cn` is generally easy. Because the standard/boundary is clear.

However, it seems harder to tell `ru` and `!ru`. The geographical boundary may be different from the cultural boundary. There are some Russian-language (or Russian-speaking users oriented) websites, while they are actually not located in/owned by Russia, maybe in Eastern Europe or Central Asian. For Russian users, maybe they should not be proxied at all. Then should we add these cultural-ru domains into `geolocation-!ru`?

For historical reasons, we have `category-ru` instead of `geolocation-ru`. I don't know what role do the Russian users want the list to play, as there's another proposal, #3424, to create `category-ru-whitelisted`. Things are a bit chaotic.

steepenergy:
> I don't know what role do the Russian users want the list to play, as there's another proposal, https://github.com/v2fly/domain-list-community/pull/3424, to create category-ru-whitelisted. Things are a bit chaotic.

To answer your question about what role `category-ru` plays for users here: it is used almost exclusively as an exclusion list to route traffic `DIRECT` (bypassing the proxy).

There are three major technical reasons why Russian users must route local sites directly rather than proxying them:

1.  **Anti-DDoS Geo-blocking:** Many Russian hosting providers use aggressive geo-blocking as a primitive defense against DDoS attacks. They simply drop all incoming connections from non-RU IP addresses.
    
2.  **Emergency Whitelisting:** During UAV (drone) or missile attacks in certain regions, local mobile operators temporarily switch their networks to a strict "whitelist" mode. In this state, access to foreign IPs and proxies is completely blocked, and only approved local resources remain accessible.
    
3.  **Mandatory VPN Blocking:** The government has required large companies (such as banks and essential services) to actively detect VPN/proxy usage and block those connections. Accessing their resources is only possible from a clean domestic residential IP.
    

Regarding the proposal #3424 to create `category-ru-whitelisted` that you mentioned: it stems from this exact situation. Because of the strict restrictions mentioned above, other users are trying to create a dedicated, guaranteed "safe" list for essential domestic services (like banking and state portals) that absolutely must bypass the proxy, separating them from general Russian media or entertainment domains.

vluisteren:
> > I was thinking whether we should have a `geolocation-!ru`?
> 
> We may need some kind of consensus or standard before creating such a list.
> 
> _I don't live in Russia. These opinions may not reflect the actual situation_
> 
> To tell whether a domain is `cn` or `!cn` is generally easy. Because the standard/boundary is clear.
> 
> However, it seems harder to tell `ru` and `!ru`. The geographical boundary may be different from the cultural boundary. There are some Russian-language (or Russian-speaking users oriented) websites, while they are actually not located in/owned by Russia, maybe in Eastern Europe or Central Asian. For Russian users, maybe they should not be proxied at all. Then should we add these cultural-ru domains into `geolocation-!ru`?
> 

I think the alpha2 region code would be from the ISO 3166-1 alpha-2.

Since `cn` didn't include the Hong Kong, Macao which are obvious under the communist party's ruling, Republic of China aka Taiwan, even its country name is China also is not part of `cn` geolocation. While East Turkestan, Tibet and Inner Mongolia in geography are not a part of China, they are still included there.

Eventually, using of the block list (black list) is too complex, I prefer recording the domain only where it is.
 
> For historical reasons, we have `category-ru` instead of `geolocation-ru`. I don't know what role do the Russian users want the list to play, as there's another proposal, #3424, to create `category-ru-whitelisted`. Things are a bit chaotic.

I would suggest we always put the region code at the last token of the file name.

I would like to say `category-ru`, `category-ir`, `category-tm` are confusing, they are bad category names, due to alpha-2 didn't tell us what it is. Since you have told me not to do to keep compatible, I won't change them here.

>Regarding the proposal https://github.com/v2fly/domain-list-community/pull/3424 to create category-ru-whitelisted that you mentioned: it stems from this exact situation. Because of the strict restrictions mentioned above, other users are trying to create a dedicated, guaranteed "safe" list for essential domestic services (like banking and state portals) that absolutely must bypass the proxy, separating them from general Russian media or entertainment domains.

If would be better to have a category called emergency-service with full name as `category-emergency-service-ru`,  and reuse `category-bank-ru`. And they can a toggle for each of them or the rest of the category with the `ru` suffix in the ruleset.

MkQtS:
@vluisteren, well, in this repo `cn` means mainland China, which is covered by the GFW. In mainland China, the domestic network is generally unhindered. Authorities (no business issues here) can take down certain website/IPs, or just a single post/article on public platforms. They are not likely to hinder a domestic website/IP, because they can take it down easily. For traffic outside, thanks to GFW and the poor international route/bandwidth, the network sucks, so we proxy them for network freedom and a better experience. This is the *clear* standard/boundary I mean.

For your proposal here, I think it is idealized and hard to achieve.

1. The Internet is generally global. Though there's GFW (or something similar) in China and some other areas. It may be hard to decide which country/area a domain belong to. A domain owned by a US company may deploy Endpoints/CDN services in China. Companies from multiple countries may work together to provide ONE service, which we need one list for that. We may have to treat some domains as global, and region code does not help then.
2. Even if we have enough region codes and a functional standard/solution to distinguish domains, we have no enough manpower to maintain that. And it may be a burden for end users to learn and use.

Actually, there are some fork or innovative projects focusing on their local region. There seems no possibility or necessity to get one universal databese for proxy usage.

vluisteren:
> 1. The Internet is generally global. Though there's GFW (or something similar) in China and some other areas. It may be hard to decide which country/area a domain belong to. A domain owned by a US company may deploy Endpoints/CDN services in China. Companies from multiple countries may work together to provide ONE service, which we need one list for that. We may have to treat some domains as global, and region code does not help then.

Knowing which AS it is or presenting a edge server would help on accessing. I care more about the enterprise ownership.

> 2. Even if we have enough region codes and a functional standard/solution to distinguish domains, we have no enough manpower to maintain that. And it may be a burden for end users to learn and use.
> 
> Actually, there are some fork or innovative projects focusing on their local region. There seems no possibility or necessity to get one universal databese for proxy usage.

https://github.com/v2fly/domain-list-community/issues/3868 is my proposal to keep the local region fork and upstream easy to sync the data. I wish it would help.

PR #3610, opened 2026-05-20
Update category-public-tracker

TripleA150:
All domains were manually checked and sorted.
Duplicate and invalid domains were removed.
New trackers were added.


TripleA150:
> dude you better not to remove `full:` because some of them host websites

Once again, I am convinced that you do not know how to read carefully.

yetanotherqubick:
Why would you add all of the (sub-)domains? For example, `opentrackr.org` clearly states that its tracker is using `tracker.opentrackr.org` domain. Other domains also have subdomains that aren't used as torrent trackers.

TripleA150:
> Why would you add all of the (sub-)domains? For example, `opentrackr.org` clearly states that its tracker is using `tracker.opentrackr.org` domain. Other domains also have subdomains that aren't used as torrent trackers.

The domains/subdomains were grouped according to the following principles:
1. If a domain has a website and a subdomain with a tracker, the full option is used.
2. If a domain does not have a website but has a subdomain with a tracker, the standard option is used to cover all subdomains that frequently change.

yetanotherqubick:
> > Why would you add all of the (sub-)domains? For example, `opentrackr.org` clearly states that its tracker is using `tracker.opentrackr.org` domain. Other domains also have subdomains that aren't used as torrent trackers.
> 
> The domains/subdomains were grouped according to the following principles:
> 
>     1. If a domain has a website and a subdomain with a tracker, the full option is used.
> 
>     2. If a domain does not have a website but has a subdomain with a tracker, the standard option is used to cover all subdomains that frequently change.

1. Well, `opentrackr.org` does have a website as do some of the others.
2. That's a wrong way. As an example `animebrowse.anidex.moe`is not a torrent tracker domain, but `anidex.moe` would include it too.

ghost:
> > dude you better not to remove `full:` because some of them host websites
> 
> Once again, I am convinced that you do not know how to read carefully.

Once again, I am convinced that you do not know how to read people's review and instead attack people LOL

PR #3664, opened 2026-06-02
category-finance: add jivara.global

Rigel-Jivara:
## What this PR does

Adds `jivara.global` as a new `data/jivara` file and includes it in `category-finance` (which is in turn included by `geolocation-!cn`).

## Why

`jivara.global` is the domain of **Jivara**, a cross-border payment platform. Its services are hosted abroad (AWS, with frontends served via CloudFront) and the platform targets users outside mainland China. The domain is currently not categorized in any list, so clients fall back to GeoIP-based routing; since the edge IPs resolve to the APAC region, the domain is frequently mis-routed as a direct connection, which breaks access for users on restricted networks.

Classifying it under `category-finance` (alongside Stripe, Wise, etc.) is semantically accurate and ensures it is correctly proxied via `geolocation-!cn`. A bare-domain entry covers all subdomains (e.g. `mp.jivara.global`, `op.jivara.global`).

Submitted by a member of the Jivara team.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/data/category-finance
+++ b/data/category-finance
@@ -18,5 +18,6 @@ include:stripe
 include:wise
 
 fxcorporate.com
+jivara.global
 mql5.com
 tradingview.com


MkQtS:
You can add `jivara.global` directly into `category-finance` instead of creating a new list for a single domain.

PR #3667, opened 2026-06-02
Create alphabet

ghost:
Add google's parent company, alphabet


MkQtS:
`google-deepmind` is already included by `google`, so no need to include it in `alphabet`.

MkQtS (approved):

PR #3680, opened 2026-06-04
Add Scholar hk category and corresponding academies

ous50:


MkQtS:
```
keyword:-polyu.hk
keyword:-polyu.edu.hk
```

Could you turn these keyword rules into specific domain type?

MkQtS (approved):

PR #3681, opened 2026-06-04
Connectivity Check: Add CN attr and more providers

ous50:

### DIFF
--- a/data/connectivity-check
+++ b/data/connectivity-check
@@ -1,5 +1,7 @@
+# Connectivity Check, default is used for checking global connectivity. use geosite:connectivity-check@cn for CHN network direct testing
+
 # Apple
-captive.apple.com
+captive.apple.com @cn # Apple captive portal default direct
 
 # Arch Linux
 full:ping.archlinux.org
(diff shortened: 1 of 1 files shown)


MkQtS:
1. The new domains should set to `full` type?
2. `conn2.oppomobile.com` can be added too
3. How about `kuketz.de` => `full:captiveportal.kuketz.de`?

ous50:
> 1. The new domains should set to `full` type?
> 2. `conn2.oppomobile.com` can be added too
> 3. How about `kuketz.de` => `full:captiveportal.kuketz.de`?

I have tested and it seems like most of `kuketz.de` domains can be used to generate_204

MkQtS (approved):

littlehappy123456:
I've been using connectivity-check rules for direct routing. So why put "www.google.com" in the list. if I change to use connectivity-check @cn for direct routing, "www.gstatic.com" is not in the list.

xkdvSrPD:
把www.google.com加入Connectivity Check是何意为
从下游一路找上来原来是你小子

PR #3682, opened 2026-06-04
Optimize CMCC and add CMI relavant rules(CMHK, CMLink, jegotrip); Add EE(for CMLink UK) and H3G(3/Three/Hutch/WindTre) 

ous50:
EE is created for CMLink UK and H3G is also included since both H3G and EE are added within a single commit.


vluisteren:
It looks good to me. I am little concern about the file `ee`, which is a very common short name. How about `eelimited` or `everythingeverywhere`

PR #3689, opened 2026-06-06
add humblebundle

peeweep:


data/category-games-!cn
MkQtS: `humblebundle.com` is already here

MkQtS (approved):

PR #3692, opened 2026-06-06
Update unity: add missing tracking domains

xiaoxizi-lyx:
### Description
This PR adds two missing Unity tracking/analytics domains to the `data/unity` file. 

**Added domains:**
* `config.uca.cloud.unity3d.com`
* `cdp.cloud.unity3d.com`

**Reason for change:**
These domains are actively used by Unity for cloud analytics, telemetry, and customer data collection (CDP). They have been placed under the `# Ads/tracking` section and properly tagged with the `@ads` attribute to ensure they can be categorized and routed correctly by users who wish to block tracking requests.

### DIFF
--- a/data/unity
+++ b/data/unity
@@ -2,5 +2,7 @@ unity.com
 unity3d.com
 
 # Ads/tracking
+config.uca.cloud.unity3d.com @ads
+cdp.cloud.unity3d.com @ads
 iads.unity3d.com @ads
 unityads.unity3d.com @ads


MkQtS:
Data source? Are they safe to block?

ghost:
> Data source? Are they safe to block?

https://discussions.unity.com/t/why-is-my-game-connecting-to-config-uca-cloud-unity3d-com-on-startup-after-recent-unity-update/184385
Turns out the answer is Unity collecting anonymous Hardware Stats from players when the game launches, which can’t be disabled with Unity Personal.

https://github.com/hagezi/dns-blocklists/issues/79
The domain is very common and present in many blocklists. It is used for in-game analytics and tracking. These analytics should actually be able to be turned off ingame. If blocking such a domain causes game functions to partially not work, this is definitely a very bad software design and should be seen as a bug of the game.

MkQtS:
Since it's problematic to block it, we'd better not to add it. You can always block it in your personal config.

PR #3712, opened 2026-06-10
steam: add more domains

yetanotherqubick:
Added `valve.net` and `dota.com`

### DIFF
--- a/data/steam
+++ b/data/steam
@@ -16,7 +16,7 @@ valve.net
 valvesoftware.com
 
 # Dota 2
-dota.com
+dota2.com
 
 # edgecast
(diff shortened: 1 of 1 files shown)


data/steam
```diff
@@ -12,8 +12,12 @@ steamserver.net
 steamstatic.com
 steamusercontent.com
 underlords.com
+valve.net
 valvesoftware.com
 
+# Dota 2
+dota.com
```
MkQtS: The domain should be `dota2.com`?

yetanotherqubick: My mistake

MkQtS (approved):

PR #3716, opened 2026-06-10
Add new analytics and performance tracking domains

xiaoxizi-lyx:
This is a resubmission and follow-up to my previously closed PR [#3692](https://github.com/v2fly/domain-list-community/pull/3692).

I sincerely apologize for the delayed response and for not providing the background information in time. I was quite busy over the past few days, and more importantly, to be responsible to the community, I spent some time testing these blocked domains on my own router to ensure that blocking them wouldn't cause any side effects or issues with the normal operation of Unity-related programs like Unity Editor or Unity Hub. After a few days of testing, I can confirm that there are absolutely no negative impacts.

These domains, which should be blocked due to ads, tracking, or telemetry, are derived directly from the **official Unity proxy exception list documentation**:
 [Defining exceptions on a Web proxy (Unity Official Manual)](https://docs.unity3d.com/Manual/ent-proxy-exception-list.html)

Below is the table summarizing the domains to be blocked and their exact purposes as described in the official documentation:

| Domain | Official Documentation Description / Actual Purpose |
| :--- | :--- |
| `analytics.cloud.unity3d.com` | Unity Hub analytics |
| `api2.amplitude.com` | Analytics for Unity Hub (via Amplitude third-party data service) |
| `config.uca.cloud.unity3d.com` | Opt-in to analytics state via remote configuration |
| `cdp.cloud.unity3d.com` | Common data platform used to collect data for analytics |
| `prd-lender.cdp.internal.unity3d.com` | Relaying data to Unity for internal analytics |
| `perf.cloud.unity3d.com` | Performance and crash reporting service |
| `perf-events.cloud.unity3d.com` | Performance event collection |
| `dashboard.unityads.unity3d.com` | Dashboard for ads and monetization |

I hope this explanation provides enough background context to support merging this PR. If there's any further information needed, please let me know and I will follow up promptly. 

Please note that I am fully aware that `domain-list-community` is just a generic list provider; I am merely helping to maintain the `Ads/tracking` category.

### DIFF
--- a/data/unity
+++ b/data/unity
@@ -2,5 +2,13 @@ unity.com
 unity3d.com
 
 # Ads/tracking
+analytics.cloud.unity3d.com @ads
+api2.amplitude.com @ads
+config.uca.cloud.unity3d.com @ads
+cdp.cloud.unity3d.com @ads
+dashboard.unityads.unity3d.com @ads
+prd-lender.cdp.internal.unity3d.com @ads
 iads.unity3d.com @ads
+perf.cloud.unity3d.com @ads
+perf-events.cloud.unity3d.com @ads
 unityads.unity3d.com @ads
--- a/data/unity
+++ b/data/unity
@@ -2,11 +2,11 @@ unity.com
 unity3d.com
 
 # Ads/tracking
+# https://github.com/v2fly/domain-list-community/pull/3716
 analytics.cloud.unity3d.com @ads
 api2.amplitude.com @ads
 config.uca.cloud.unity3d.com @ads
 cdp.cloud.unity3d.com @ads
-dashboard.unityads.unity3d.com @ads
 prd-lender.cdp.internal.unity3d.com @ads
 iads.unity3d.com @ads
 perf.cloud.unity3d.com @ads


ghost:
You could block them yourself; `domain-list-community` is just a generic list provider and doesn't have "blacklist"

xiaoxizi-lyx:
> You could block them yourself; `domain-list-community` is just a generic list provider and doesn't have "blacklist"

Sorry, perhaps my wording was a bit off; the blacklist was just a metaphor. What I meant was, since there's already a `Ads/tracking` category, what's wrong with adding domains related to `Ads/tracking` to it?

MkQtS:
I removed `dashboard.unityads.unity3d.com @ads` because it will be overrided by `unityads.unity3d.com @ads`

---

The information should be enough now.

MkQtS (approved):

anivaros:
Why does Amplitude included in the Unity ADS list?!

MkQtS:
> These domains, which should be blocked due to ads, tracking, or telemetry, are derived directly from the **official Unity proxy exception list documentation**:
> [Defining exceptions on a Web proxy (Unity Official Manual)](https://docs.unity3d.com/Manual/ent-proxy-exception-list.html)

@anivaros, `api2.amplitude.com` came from the link above, which is described as `Analytics for Unity Hub`.

The domain does not belong to Unity. We can remove it if necessary.

anivaros:
@MkQtS If possible, that would be great, thank you!

MkQtS:
@xiaoxizi-lyx 

Any objections?

xiaoxizi-lyx:
@MkQtS

I'm fine with removing `api2.amplitude.com` from the `unity` file specifically, since it's not actually a Unity-owned domain — Amplitude is a general-purpose third-party analytics/tracking provider that's used far beyond just Unity Hub, so it doesn't really belong in a Unity-specific list.

However, I don't think it should be removed from the project entirely. The fact that "Amplitude isn't Unity-exclusive" doesn't change the fact that `api2.amplitude.com` is genuinely a tracking domain — which is exactly the point you made in #3990 about tracking often being more invasive than ads themselves. Dropping it from the whole project would mean losing that protection for every product that uses Amplitude, not just Unity ones.

So my suggestion would be: relocate it rather than delete it. If there's already a suitable general list for third-party trackers, we could move it there and tag it `@ads`. If not, maybe it's worth creating a dedicated `amplitude` entry (similar to how other cross-product trackers like Google Analytics are handled) so it stays covered under `Ads/tracking` without being misattributed to Unity.

Let me know what makes sense on your end — happy to submit a follow-up PR to move it if that's easier.

MkQtS:
You can create a new list of `amplitude`, and then include it in `category-ads-all`, like `sentry`. Or if there's only one domain, `amplitude.com`, then just add it directly in `category-ads-all`.

anivaros:
This is a very strange decision, because before this pr two months ago, there was no amplitude in any list at all.

It would be logical to simply create an Amplitude list so that if someone needs to, they can do something with their domains -- block or something else. So and other analytics services.

PR #3728, opened 2026-06-12
Update category-ru

GoodDuckYy:
Add accounts.google.ru to category-ru

### DIFF
--- a/data/category-ru
+++ b/data/category-ru
@@ -72,6 +72,7 @@ dom.ru
 
 # Other services
 3ebra.net          # Zebra Net Bureau, web development agency
+accounts.google.ru # Google Account service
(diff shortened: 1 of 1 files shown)


MkQtS:
1. There is `[domain:]ru` in `category-ru` (from `data/tld-ru`), which matches all `*.ru` domains.
2. When building the `dlc.dat`, redundant subdomains are trimed. So `*.ru` rules in `category-ru` will not appear in the generated binary file.
3. There is some `*.ru` domains in `data/category-ru`, and they do not contribute to the `dlc.dat`. They are there only for users to view or maybe for downstream projects with different logic.
4. I don't see any value to add a subdomain of `google.ru` into `data/category-ru`

PR #3748, opened 2026-06-15
Add full:gist.github.com to github

astramare:
### Description
Add `full:gist.github.com` explicitly to the `github` domain list.

### Reason
While `github.com` generally covers subdomains via suffix matching in some environments, certain strict downstream routing applications and compiled binary rule-sets (such as strict Radix-tree based compilers) fail to correctly route or generate nodes for `gist.github.com` when it's not explicitly declared. 

Since Gist is a core and essential service of the GitHub platform, adding it explicitly as a `full:` domain acts as a necessary high-precision patch. This ensures robust routing and consistent network policy enforcement without introducing generic subdomain redundancy.

### DIFF
--- a/data/github
+++ b/data/github
@@ -26,6 +26,7 @@ octocaptcha.com
 opensource.guide
 repo.new
 thegithubshop.com
+gist.github.com
 
 full:github-api.arkoselabs.com
 


database64128:
The reasoning reads AI generated nonsense to me.

PR #3755, opened 2026-06-17
Update JetBrains

windchargerj:
Add new JetBrains domains for CDN and cache.


MkQtS:
The two domains do not resolve to IP address in cn.

PR #3805, opened 2026-07-04
移除 tailscale 数据中的 ts.net，并新增 tailscale-magicdns 文件以包含 ts.net

zyzh2002:
Open a separate file for tailnet's magicdns to achieve finer-grained routing control.

### DIFF
--- a/data/tailscale
+++ b/data/tailscale
@@ -1,3 +1,2 @@
 tailscale.com
 tailscale.io
-ts.net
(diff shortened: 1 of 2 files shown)


MkQtS:
Is this necessary? Tailscale users can add a special rule for the single domain `ts.net`. Why do we need `geosite:tailscale-magicdns`?

PR #3814, opened 2026-07-07
add aux

peeweep:


MkQtS (approved):

ivanstus:
After this commit repository cannot be used on Windows since `aux` is invalid filename.

MkQtS:
Let's rename it to `auxgroup`...

peeweep:
Thanks for report it @ivanstus  and thanks for fix it @MkQtS

PR #3816, opened 2026-07-08
Update & sort apple-related categories

direnquanbuKO:
1. In `apple` category, I sorted the list alphabetically according to the comment titles, plus added some domains that have access points in China Mainland for Apple Podcasts
2. In `apple-tvplus` and `itunes`, I also updated some domains

## Proposal

I think we can extract the `apple` category into different categories, like `apple-music`, `apple-podcasts` and so on, because in this category there are so many domains with many services, especially in the following lines

https://github.com/v2fly/domain-list-community/blob/fcb888537ae69314c53b9ffae7dcc594fddfd8c6/data/apple#L752-L775

They are mixed in a category but with several different services, I think it's better to put them into different categories

### DIFF
--- a/data/apple
+++ b/data/apple
@@ -772,7 +772,7 @@ full:km.support.apple.com @cn
 full:maps.apple.com @cn
 full:musicstatus.music.apple.com @cn
 full:play.music.apple.com @cn
-full:podcasts.apple.com
+full:podcasts.apple.com @cn
 full:speedysub.music.apple.com @cn
 full:www.support.apple.com @cn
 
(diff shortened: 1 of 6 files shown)


MkQtS:
1. In `apple-tvplus`, there's `[domain:]tv.apple.com`, so `full:amp-account.tv.apple.com` and `full:buy.tv.apple.com` are redundant.
2. For your proposal, there're many sections in `data/apple`, some of them may deserve independent lists, some not. Please only add new lists which serve for actual needs.

MkQtS:
Please include apple-music and apple-podcasts in category-entertainment

MkQtS (approved):

PR #3824, opened 2026-07-10
category-electronic-cn: fix espressif

vluisteren:
Since there is not a category file for the electronic, puting those domains out of the company file didn't present the right ownership.

Besides, may I suggest we add a new attribute for those domains with a edge(local) server, marking them with a region code is not right.
We can introduce a attribute likes `@edge=cn`

### DIFF
--- a/data/geolocation-!cn
+++ b/data/geolocation-!cn
@@ -321,7 +321,7 @@ wiki.gg
 
 # Others
 include:avaxhome
-include:espressif
+include:espressif @!cn
 include:familymart
 include:fzdm
 include:hkedcity
--- a/data/category-electronic-cn
+++ b/data/category-electronic-cn
@@ -20,11 +20,7 @@ bouffalolab.com
 cxmt.com
 
 # 乐鑫信息科技
-#include:espressif
-esp8266.cn
-esp8266.com.cn
-espressif.cn
-espressif.com.cn
+include:espressif @-!cn
 
 # 华秋电子
 eda.cn
--- a/data/espressif
+++ b/data/espressif
@@ -1,3 +1,7 @@
-espressif.com
-esp32.com
-esp8266.com
+esp32.com @!cn
+esp8266.cn
+esp8266.com @!cn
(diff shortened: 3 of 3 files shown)


MkQtS:
https://github.com/v2fly/domain-list-community/blob/086c4b2901df3e88420830a37cd86daa376c3e14/data/geolocation-!cn#L324

`espressif` is included in `geolocation-!cn`, that's why I didn't add these espressif-cn domains into  `data/espressif`

vluisteren:
> https://github.com/v2fly/domain-list-community/blob/086c4b2901df3e88420830a37cd86daa376c3e14/data/geolocation-!cn#L324
> 
> `espressif` is included in `geolocation-!cn`, that's why I didn't add these espressif-cn domains into `data/espressif`

Then it is better to remove it from `geolocation-!cn`. I am developing a inspect tools, during its figure out the high frequency domain name(or multiple tld variants) inspection, it wound this one.

My intention is keep the right enterprise hierarchy keep.

MkQtS (approved):

data/geolocation-!cn
```diff
@@ -321,7 +321,7 @@ wiki.gg
 
 # Others
 include:avaxhome
-include:espressif
+include:espressif @!cn
```
vluisteren, line 324: I am afraid it is the first  and the very statement using such limitation selection option.
From the syntax, it is correct, it is just describe it twice. 

We have known those domains don't have a Chinese access server, should we tell people again in the geolocation file?

MkQtS: I added this to keep the consistency for the espressif domains in `geolocation-!cn`, since they were initially added to be in `geolocation-!cn`. This doesn't mean I'm going to introduce/apply significant changes like #3265

If you like, you can open another PR to add some notes/statements for this.

vluisteren: I would like to do so after I read others idea at https://github.com/v2fly/domain-list-community/issues/28

PR #3825, opened 2026-07-10
Add sentry tracker domain

yobarerukoto:
Reference: https://docs.sentry.io/security-legal-pii/security/ip-ranges/#event-ingestion

### DIFF
--- a/data/sentry
+++ b/data/sentry
@@ -1,2 +1,8 @@
 sentry.io
 sentry-cdn.com
+
+# tracking
+regexp:^o\d+\.ingest\.sentry\.io$ @ads
+regexp:^o\d+\.ingest\.us2\.sentry\.io$ @ads
+regexp:^o\d+\.ingest\.de\.sentry\.io$ @ads
+regexp:^o\d+\.ingest\.us\.sentry\.io$ @ads


MkQtS:
There is `[domain:]sentry.io` which matches all `*.sentry.io` domains. All regex rules you added here are redundant.

yobarerukoto:
> There is `[domain:]sentry.io` which matches all `*.sentry.io` domains. All regex rules you added here are redundant.

It is necessary to distinguish regular domains from tracker domains. @MkQtS

MkQtS:
Generally, all `*.sentry.io` domains are used for tracking. And the performance of regexp rules is bad.

yobarerukoto:
Sentry also hosts its website and documentation under the same domain, such as `docs.sentry.io`.
If this is being rejected because of regex performance concerns, would it be better for just using a specific domain like `ingest.sentry.io`?

MkQtS:
Well, what's your purpose then?

Currently, sentry is not included in `category-ads` or `category-ads-all`, even if you added `@ads` to some domains in `data/sentry`. We can include sentry directly in `category-ads-all`.

PR #3840, opened 2026-07-12
remove invalid top level domain records

vluisteren:
Can't find them in the IANA list:
.adobe
.intel
.goo (from google)
.konami
.xn--mgba3a4fra (Iran's Desired Variant String, not allocated yet)
.ipp (from adobe-activation)
.nowoa (from adobe-activation)
.ntp (from adobe-activation)
.psk (from category-ru, it may be only valid in Russia ISPs)

Suppose they are only valid in a local area network, not for public resolver:
.amc (from category-bank-ru, it looks like a package name)
.sipus (from category-stun)
.prg (from category-stun)
.local (from category-public-tracker, record full:retracker.local,
it is a Russian intranet domain)

Those TLDs should be put into either private file or mark them excluded from the v2ray rule

### DIFF
--- a/data/adobe-activation
+++ b/data/adobe-activation
@@ -39,11 +39,7 @@ full:lm.licenses.adobe.com
 full:lmlicenses.wip4.adobe.com
 full:na1r.services.adobe.com
 full:ood.opsource.net
-full:practivate.adobe
 full:practivate.adobe.com
-full:practivate.adobe.ipp
-full:practivate.adobe.newoa
-full:practivate.adobe.ntp
 full:wip.adobe.com
 full:wip1.adobe.com
 full:wip2.adobe.com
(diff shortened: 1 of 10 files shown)


MkQtS:
Generally fine. For the `psk` in `category-gov-ru`, maybe it's desired/necessary for Russian users. I'm not sure whether we should remove it.

Ping some ru contributors for help. (Sorry for the interruptions) @wgos @TripleA150 @mazixs @yetanotherqubick @ACherepkov1989

yetanotherqubick (approved):
Never heard of `.psk` to be honest. `domain:alfabank.oavdo.amc` can be removed too — I don't know why I added it in the first place.

MkQtS (approved):

PR #3849, opened 2026-07-14
strip existed enterprise owned domain from category

vluisteren:
TODO:
category-game-platforms-download
data/category-game-accelerator-cn

I don't know what I should do for them, for example, the resources fqdn for the xbox, the xbox itself is a product name under the enterprise mircosoft

### DIFF
--- a/data/category-ai-ru
+++ b/data/category-ai-ru
@@ -2,9 +2,7 @@
 aiesa.ru
 
 # Yandex Alice
-alice.yandex.net
-alice.yandex.ru
-alicepro.yandex.ru
+include:yandex-ai
 
 # CopyMonkey (AI copywriter)
 copymonkey.app
--- a/data/category-doh-cn
+++ b/data/category-doh-cn
@@ -11,12 +11,7 @@ httpdns.n.shifen.com @ads
 httpsdns.baidu.com @ads
 
 # Huawei Cloud
-httpdns-browser.platform.dbankcloud.cn @ads
-httpdns.c.cdnhwc2.com @ads
-httpdns.huaweicloud.com @ads
-httpdns.platform.dbankcloud.cn @ads
-httpdns.platform.dbankcloud.com @ads
-httpdns1.cc.cdnhwc5.com @ads
+include:huawei-doh
 
 # iqiyi
 dns.iqiyi.com @ads
@@ -43,9 +38,7 @@ httpdns.push.oppomobile.com @ads
 httpdns.ocloud.oppomobile.com @ads
 
 # Tencent
-dns.weixin.qq.com @ads
-dns.weixin.qq.com.cn @ads
-httpdns.kg.qq.com @ads
+include:tencent-doh
 
 # HeyTap
 httpdns.ocloud.heytapmobi.com @ads
--- a/data/geolocation-cn
+++ b/data/geolocation-cn
@@ -31,8 +31,8 @@ qchannel03.cn
 # 神策数据
 sensorsdata.cn
 
-# category-httpdns-cn is mainly for advertising purpose
-include:category-httpdns-cn
+# category-doh-cn is mainly for advertising purpose
+include:category-doh-cn
 
 # 号码认证
 # include:category-number-verification-cn
--- a/data/huawei
+++ b/data/huawei
@@ -1,5 +1,6 @@
 include:huawei-dev
 include:huaweicloud
+include:huawei-doh
 
 abhouses.com
 agconnect.link
--- a/data/huawei-doh
+++ b/data/huawei-doh
@@ -1,0 +1,6 @@
+httpdns-browser.platform.dbankcloud.cn @ads
+httpdns.c.cdnhwc2.com @ads
+httpdns.huaweicloud.com @ads
+httpdns.platform.dbankcloud.cn @ads
+httpdns.platform.dbankcloud.com @ads
+httpdns1.cc.cdnhwc5.com @ads
--- a/data/tencent
+++ b/data/tencent
@@ -4,6 +4,8 @@ include:tencent-dev
 include:tencent-games
 include:tencent-tme
 include:yuewen
+# Tencent Wechat ads
+include:tencent-doh
 
 qq.xn--fiqs8s # qq.中国
 xn--r70as2s.xn--fiqs8s # 腾讯.中国
--- a/data/tencent-doh
+++ b/data/tencent-doh
@@ -1,0 +1,3 @@
+dns.weixin.qq.com @ads
+dns.weixin.qq.com.cn @ads
+httpdns.kg.qq.com @ads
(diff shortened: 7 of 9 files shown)


MkQtS:
1. `httpdns` is differnet from standard DoH, from protocol to usage. And the rename may break existing configurations.
2. You don't need to include every domain into the entity it belongs to. Lists like `tencent-doh` do not make much sense. `geosite:tencent` contains `[domain:]qq.com[.cn]` and can match these domains in `tencent-doh`. We have 1k+ lists in `data/`. Such new lists are not actually needed by users, and make things complex/confusing.

   To add a domain to an entity and a category, there's a new syntax, affiliation, but currently we should avoid using it to keep downstream projects working. You can write the domains twice in two lists. For the situation here, adding these subdomains into their parent entity makes no sense as they are *redundant*. Creating middle list and double inclusions works, but as said, it make things complex/confusing, so only do it when we *need* the middle list.

Generally, these domains are in appropriate categories. No need to be obsessed with classification.

vluisteren:
> 1. `httpdns` is differnet from standard DoH, from protocol to usage. And the rename may break existing configurations.

I see, since I don't live in China, I don't know what it is. If it is a different protocol, I should not rename it.

> 2. You don't need to include every domain into the entity it belongs to. Lists like `tencent-doh` do not make much sense. `geosite:tencent` contains `[domain:]qq.com[.cn]` and can match these domains in `tencent-doh`. We have 1k+ lists in `data/`. Such new lists are not actually needed by users, and make things complex/confusing.
>    To add a domain to an entity and a category, there's a new syntax, affiliation, but currently we should avoid using it to keep downstream projects working. You can write the domains twice in two lists. For the situation here, adding these subdomains into their parent entity makes no sense as they are _redundant_. Creating middle list and double inclusions works, but as said, it make things complex/confusing, so only do it when we _need_ the middle list.
> 
> Generally, these domains are in appropriate categories. No need to be obsessed with classification.

The reason that I want to do it, I read an article about throttling in the China network, they made an example that if you were querying a tencent's service in a alibaba's dns server, that dns server would return the worst result.

It is ever worse in the ROK (south korea) that accessing Japanese servers is much fast than the other demostic provider due to SPNP regulatory.

Also from what I know the net neutrality is only enforced in a few countries.

Is there a topic how we can improve the hierarchy of those rules? The AI translator only gave me a confusing result, @MkQtS  I would like close this MR and move to there.

MkQtS:
> The reason that I want to do it, I read an article about throttling in the China network, they made an example that if you were querying a tencent's service in a alibaba's dns server, that dns server would return the worst result.

DNS poisoning exists, but it's mostly for global websites blocked by GFW. All companies in China have to comply, not related to business competition.

Sometimes, DNS poisoning is conducted by the local ISP, to hijack some websites and insert ads. Most business companies are against this. This is one of the reasons that these famous public national DNS (and httpdns) in China, like alidns and tencent dns, exist. Setting a single domestic DNS for domestic websites/apps generally works, even if when alidns is serving for tencent domains.

> Is there a topic how we can improve the hierarchy of those rules?

Probably #28. (It's mainly in Chinese and actually no active disscussion)

vluisteren:
> > The reason that I want to do it, I read an article about throttling in the China network, they made an example that if you were querying a tencent's service in a alibaba's dns server, that dns server would return the worst result.
> 
> DNS poisoning exists, but it's mostly for global websites blocked by GFW. All companies in China have to comply, not related to business competition.
> 

It is just the rumor I heard. There were many edge servers in different provinces, while a service is hosted at alibaba network, if you were querying it from a tencent's public resolver, it would not return the optimized result. The result could the worst line for the requester(ex. the requester is querying from telecomm A, state B, the resolver would tell you an edge server at telecomm D, state E).

> Sometimes, DNS poisoning is conducted by the local ISP, to hijack some websites and insert ads. Most business companies are against this. This is one of the reasons that these famous public national DNS (and httpdns) in China, like alidns and tencent dns, exist. Setting a single domestic DNS for domestic websites/apps generally works, even if when alidns is serving for tencent domains.
> 

Throttling is so common could be found all around the world. Also it can save some moneys on interanet access (for the cloud service).

I also heard even for the same telecomm group in China, sub branches of would charge peer usages. So I want to make the ownership clear.

The reason why the first level of the subdomain is not canonical, we need to handle division or product line migration. a bu1.group1 may not owned by the enterprise foo since enteprise bar bought it, the acquiring contract may request the enterprise foo to keep those domains for a period (also email address to keep in touch with old customers)

> Is there a topic how we can improve the hierarchy of those rules?
> 
> Probably #28. (It's mainly in Chinese and actually no active disscussion)

I would translate my ideas back to Chinese.

MkQtS:
To a certain extent, these 'rumor' are real. But it seems a more vague and general topic.

---

Lists of entities should represent the correct ownership. But for `category-*` lists, sometimes, it's  more about the topic/usage rather than the ownership.

In my personal opinion, current plain data structure is not quite suitable for presenting precise/complex relationship. To pursue some kind of cannonicalization, we may get a suitable data structure first, maybe mutiple folders in `data/` and relational databases... But it seems far from our actual needs/manpower.

---

I'll close this PR. You can still comment here if you like.

vluisteren:
> To a certain extent, these 'rumor' are real. But it seems a more vague and general topic.
> 
> Lists of entities should represent the correct ownership. But for `category-*` lists, sometimes, it's more about the topic/usage rather than the ownership.
> 
> In my personal opinion, current plain data structure is not quite suitable for presenting precise/complex relationship. To pursue some kind of cannonicalization, we may get a suitable data structure first, maybe mutiple folders in `data/` and relational databases... But it seems far from our actual needs/manpower.
> 
> I'll close this PR. You can still comment here if you like.

Would you like to forward the description of that rumor to https://github.com/v2fly/domain-list-community/issues/28 ?
That is the reason I would like to make everything in the right ownership.

MkQtS:
> Would you like to forward the description of that rumor to #28 ? That is the reason I would like to make everything in the right ownership.

You can do it as you want.

vluisteren:
> > Would you like to forward the description of that rumor to #28 ? That is the reason I would like to make everything in the right ownership.
> 
> You can do it as you want.

Sorry, I am afraid I don't know whether I can explain the necessary of this clearly beyond the cloud part, and I can't tell whether the AI translation work correctly.

If you can take a time on this, I would appreciate it.

MkQtS:
> I can't tell whether the AI translation work correctly.

I think the AI translation is good. And you can post in English as we can read English by ourselves or translators.

TBH, I don't quite agree with your logic between business-related DNS throttling with the ownership and the necessarity of renaming.

My opinions are:

   1. Lists of entities should represent the correct ownership. But for `category-*` lists, sometimes, it's more about the topic/usage rather than the ownership.
   2. If we deleted/renamed an existing list, their clients fails and we get complaints. So we need to be cautious and only do so when it's safe/necessary.

PR #3852, opened 2026-07-15
Update downloads.openwrt.org

tooadstool:


MkQtS:
`[domain:]openwrt.org` already matches `downloads.openwrt.org`. Why do we need the *redundant* subdomain?

PR #3859, opened 2026-07-19
Added formula1 domains

skoif:
Added Formula 1 domains.

Using the "full:" prefix because F1 have multiple subdomains

### DIFF
--- a/data/category-entertainment
+++ b/data/category-entertainment
@@ -31,6 +31,7 @@ include:discoveryplus
 include:disney
 include:dlsite
 include:ficbook
+include:formula1
 include:gimy
 include:goodreads
 include:hamivideo
--- a/data/formula1
+++ b/data/formula1
@@ -1,4 +1,3 @@
 formula1.com
 f1.com
 f1tv.com
-f1-live.com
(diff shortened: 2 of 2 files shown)


MkQtS:
If you want to match subdomains, then you should use the default `domain:` type. (Just remove the `full:` prefix)

`full:f1.com` matches the exact domain `f1.com`, `*.f1.com` is not matched.

MkQtS:
No offense. Why did you remove `formula1.com.edgesuite.net`? It seems fine to be added in the list.

skoif:
> No offense. Why did you remove `formula1.com.edgesuite.net`? It seems fine to be added in the list.

That domain actually carried over from my home router’s configuration. It turns out the domain doesn’t actually exist. I checked it with nslookup and it doesn’t resolve, so there’s no point keeping it in the list.

MkQtS:
Please include the new list in `category-entertainment`

skoif:
> That domain actually carried over from my home router’s configuration. It turns out the domain doesn’t actually exist. I checked it with nslookup and it doesn’t resolve, so there’s no point keeping it in the list.

Same for f1-live.com, will be removed in the next commit

MkQtS (approved):

PR #3871, opened 2026-07-25
Exclude tld-ru

TripleA150:
Exclude tld-ru from category-ru for selective domain coverage

### DIFF
--- a/data/category-ru
+++ b/data/category-ru
@@ -1,11 +1,5 @@
-include:tld-ru
 ru.com
 ru.net
-
-# - There is `[domain:]ru` in this list (from tld-ru), which matches all `*.ru` domains
-# - Redundant subdomains (full/domain type without attr) will be trimed by main.go
-#   In this list, `*.ru` domains without attr are redundant for the program and will not contribute to `geosite:category-ru`
-# - You can still add `*.ru` domains here if you want
 
 # Subcategories
 include:category-ai-ru


MkQtS:
This seems a signaificant change. It's better to have some kind of consensus.

For users in Russia, welcome to share your ideas.

TripleA150:
> This seems a signaificant change. It's better to have some kind of consensus.
> 
> For users in Russia, welcome to share your ideas.

Yes, that would be better. 
My idea:
Previously, TLDs were needed because category-ru was insufficiently populated. Now that most companies have been added to category-ru, in my opinion, TLDs are just getting in the way, covering the entire zones rather than specific company domains.

ACherepkov1989:
This change will break existing routing rules. The best solution is to create a software rule that will block this "tld-ru".
`"BlockSites": ["geosite:tld-ru"]`

TripleA150:
As it stands, the added "tld-ru" list overrides (excludes) all subdomains within "category-ru" during ".dat" file generation; consequently, your proposed approach looks illogical, as it would block (remove) the entire "ru" zone rather than just specific TLDs.

An inefficient software solution would be to gather all lists and domains - excluding "tld-ru" and generate a custom "category-ru".

Instead, I propose creating a separate list that combines these two categories ("category-ru" and "tld-ru") for example, named "ru," following the pattern used for "cn."

If you have better solutions, please suggest them. I'd be happy to hear them :)

@MkQtS @ACherepkov1989

yetanotherqubick:
I think it should stay the way it is now. Absolute majority of domains in `.ru` TLD zone are Russian websites that use infrastructure located in Russia. If some of them are blocked in Russia, then one can proxy these explicitly. I can remember only a few `.ru` websites that have this kind of issues.

ACherepkov1989:
@TripleA150, there's a suitable topic for your solution: category-ru-whitelisted #3424.

TripleA150:
Ok. Thanks guys for the discussion!

PR #3881, opened 2026-07-28
update kubernetes category: add pkgs.k8s.io

KnYaZ-95:
Add pkgs.k8s.io (when you can download packages)


MkQtS:
`[domain:]k8s.io` matches `pkgs.k8s.io`, so this is redundant.

PR #3903, opened 2026-08-03
Add Activision

TripleA150:


MkQtS:
According to the [wikipedia](https://en.wikipedia.org/wiki/Activision_Blizzard), Activision Blizzard currently includes three operating units: Activision, Blizzard Entertainment and King.

I think we can create `data/activision-blizzard` and `data/activision` to represent the correct ownership (the King seems too small to create a seperate list). Like:

```
# data/activision-blizzard
include:activision
include:blizzard
king.com
```

BTW, we should probably not include ctivision-blizzard in microsoft/xbox. Because some Chinese users are used to make all microsoft traffic direct.

TripleA150:
@MkQtS done :)

MkQtS (approved):

PR #3912, opened 2026-08-04
main.go: support to trim subdomains with attrs

MkQtS:


database64128 (approved):

MkQtS:
Off topic. The ci script to print the diff of `dlc.dat_plain.yml` (I introduced in #3888) always uses newer main.go, so it can not show dlc.dat changes caused by main.go changes...

PR #3916, opened 2026-08-05
Add ncbi.nlm.nih.gov to category-scholar-!cn

y1zhou:
Direct connections to NCBI is not stable, and Google reCAPTCHA often blocks access.

### DIFF
--- a/data/category-scholar-!cn
+++ b/data/category-scholar-!cn
@@ -130,6 +130,7 @@ medrxiv.org
 morganclaypool.com
 nature.com
 naturemag.org
+ncbi.nlm.nih.gov
 ncl.edu.tw
 nejm.org
 neurology.org


MkQtS:
Maybe it's better to add `nlm.nih.gov`?

y1zhou:
> Maybe it's better to add `nlm.nih.gov`?

Thanks! That may cover additional resources listed in <https://www.nlm.nih.gov/>; if we are okay with `category-scholar-!cn` covering more than just scientific publications then I think `nlm.nih.gov` is fine. I'll leave it for the maintainers to decide.

MkQtS:
Fine, I'll add `ncbi.nlm.nih.gov` for now. If anyone thinks `nlm.nih.gov` is necessary, we can update then.

MkQtS (approved):

PR #3936, opened 2026-08-10
Update escapefromtarkov

Kortart:


MkQtS:
`eft.com` is not related with the game. See https://www.etf.com/about-us

PR #3938, opened 2026-08-10
add api.github.com entry to github-copilot data

F50Ci37Y:
When you use the **Generate Commit Message** button in VS Code, VS Code sends a request to `api.github.com`, and this address was not included in `github-copilot`.

You can see the logs in the screenshot below. At first, the request failed because this address is blocked in my country. However, after adding it to the proxy, the request went through successfully.
Screenshot_20260810_193640


MkQtS:
https://github.com/v2fly/domain-list-community/pull/3542#issuecomment-4387513206

PR #3942, opened 2026-08-11
Add Yandex speed test domains to category-speedtest

vladon:


MakostaDev:
Yandex has domains in various zones. For example, `.tr` and `.kz`. So [internet.yandex.kz](https://internet.yandex.kz) and [internet.yandex.tr](https://internet.yandex.tr) work.

[internet.ya.ru](https://internet.ya.ru) also works.

Check which other domains are working

PR #3952, opened 2026-08-12
Update 7tv

AlexStank:
Source:
image


EpLiar:
`7tv.app` is already included the sub-domain like `cdn.7tv.app` and `api.7tv.app`

PR #3955, opened 2026-08-14
Add lingxing ziniao cn

alanyilun-sketch:

### DIFF
--- a/data/lingxing
+++ b/data/lingxing
@@ -1,0 +1,1 @@
+lingxing.com &geolocation-cn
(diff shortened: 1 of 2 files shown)


MkQtS:
1. Affiliation should not be used for now to keep downstream projects work.
2. Creating new lists for a single domain is not encouraged
3. Affiliation is used to reduce the number of files in `data/`, not to increase it. Why not add the domains directly in `data/geolocation-cn` or include the new lists?

PR #3964, opened 2026-08-17
Create new iranintl data file and add iranintl to category-media

F50Ci37Y:


MkQtS:
Data source? Why `full:daq1pvcuhj77l.cloudfront.net` and `full:live.livetvstream.co.uk` should be added in the list of `iranintl`? `full:live.livetvstream.co.uk` is not likely to be a domain that belongs to the site or exclusively serves for it.

F50Ci37Y:
> Data source? Why `full:daq1pvcuhj77l.cloudfront.net` and `full:live.livetvstream.co.uk` should be added in the list of `iranintl`? `full:live.livetvstream.co.uk` is not likely to be a domain that belongs to the site or exclusively serves for it.

For `full:daq1pvcuhj77l.cloudfront.net`, the reason is that it is used as a CDN by Iran International. Here is a example link:
https://daq1pvcuhj77l.cloudfront.net/1/d0d72849af2e41c48675e89a93484507/59890cdf-9f4e-46bb-b8dd-2f6989f4b86b_thumb_001_HD.jpg

For `full:live.livetvstream.co.uk`, it is used for the live stream. If you open the site, you can see the live stream on the right side:

https://www.iranintl.com/en

PR #3987, opened 2026-08-24
Add iQIYI risk-control domain 20130123.com

boheastill:
## What

Add `20130123.com` to `data/iqiyi`. It is iQIYI's risk-control / anti-fraud token service, used by their web player.

## Evidence

Since around **2026-08-24**, iQIYI's PC web player (`www.iqiyi.com`) requires a token from this domain before it renders bullet comments (danmaku):

- The player page requests `https://zx.20130123.com/jsapi/token/iqiyi` with `Origin: https://www.iqiyi.com`; a normal response is `{"ret":0,"msg":"","token":"...","expire_at":...}`.
- The legacy danmaku endpoint `bullet.iqiyi.com` now returns NXDOMAIN from the `iqiyi.com` authoritative NS (SOA serial `2026082415`), i.e. the danmaku stack was migrated to new domains around that date.
- If this token request fails, video playback still works but **no danmaku is rendered**; the player retries the endpoint about every 5 seconds.
- `zx.20130123.com` resolves (via AliDNS, `223.5.5.5`) into ranges that also serve other iQIYI infrastructure — observed answers rotate between China Mobile Guangdong (`120.241.137.95`, inside `120.241.0.0/16`) and China Telecom Guangdong (`163.177.28.186`). I have not observed this domain being used by any service other than the iQIYI player.
- WHOIS registrant is privacy-protected, but the domain was registered on 2013-01-22 (matching iQIYI's era) and uses AliDNS nameservers (`vip1/vip2.alidns.com`). Hence the behavioral evidence above.

## Impact

Routing setups that send CN traffic direct based on `geosite-cn` currently miss this domain, so the risk-control endpoint goes through their foreign exit — where it does not accept connections (TCP timeout). Since the migration, danmaku silently disappears for all such users while videos keep playing. Adding the domain restores danmaku.

## Scope

Single-file change (`data/iqiyi`), one line, alphabetical order preserved (`20130123.com` sorts before `71.am`).

### DIFF
--- a/data/iqiyi
+++ b/data/iqiyi
@@ -1,3 +1,4 @@
+20130123.com
 71.am
 iq.com @!cn
 iqiyi.com
@@ -21,4 +22,4 @@ ifacelog.iqiyi.com @ads
 mbdlog.iqiyi.com @ads
 msg.video.qiyi.com @ads
 msg2.video.qiyi.com @ads
-policy.video.iqiyi.com @ads
+policy.video.qiyi.com @ads


MkQtS:
1. Why did you change `policy.video.iqiyi.com @ads` into `policy.video.qiyi.com @ads`? The later one resolves to NXDOMAIN
2. The domain `20130123.com` may have certain cooperation with iQIYI. But it belongs to `深圳市掌信传媒`. Besides, `https://www.20130123.com/map.html` is a page of Baidu map. It's better to be added in `geolocation-cn` instead.

boheastill:
Thanks for landing this in `geolocation-cn` (#4003) — exactly the right place, and the registrant annotation is a nice touch. Confirmed working on our side. Also fair point on the `policy.video.qiyi.com` change — an unintended drive-by edit in our diff that shouldn't have been there; lesson logged. Appreciate the quick turnaround!

PR #3993, opened 2026-08-25
feat(newgrounds): add newgrounds.com and ngfiles.com to the domain list

F50Ci37Y:


MkQtS:
Maybe it should be included in `category-entertainment` instead?

F50Ci37Y:
> Maybe it should be included in `category-entertainment` instead?

Newgrounds.com was already under the `category-porn` category. I only created a separate category for it because it has two domains.

A significant portion of the site is NSFW, and only a small portion of its content is non-adult.

MkQtS (approved):

PR #4001, opened 2026-08-27
`category-ai-cn`: add more domains

MakostaDev:
When visiting [chat.z.ai,](https://chat.z.ai) requests are also made to [chatglm.site](https://chatglm.site) and [space-z.ai](https://space-z.ai). Adding both domains to the Zhipu (智谱) group.

### DIFF
--- a/data/category-ai-cn
+++ b/data/category-ai-cn
@@ -54,6 +54,8 @@ tiangong.cn
 # 智谱
 bigmodel.cn
 chatglm.cn
+chatglm.site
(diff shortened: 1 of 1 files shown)


MkQtS:
The domains should belong to Zhipu (智谱). But they seem to be located in HongKong, and have no ICP licience (备案). Maybe they are serving for oversea market. Maybe `chat.z.ai` use them to check user's IP? I don't know whether it's proper to add them here.

PR #4075, opened 2026-09-19
Add Melli and Sepah bank's recently introduced domains

hosni:
Because of sanctions on issuing SSL certificates, Iranian banks are introducing new domains for their online services


Bassamx93 (commented):
@chaifeng

